Issue #26, opened 2015-03-17
Reconnections do not correctly initialise


stuartpullinger:
When the ssm receiver loses its connection, it attempts to reconnect. It attempts to connect to each broker in the brokers list. After the first reconnection attempt, the ssm is not correctly initialised with the host and port of the second broker.

2015-03-13 18:03:12,652 - ssmreceive - ERROR - Connection lost.
2015-03-13 18:03:12,652 - ssm.ssm2 - INFO - SSM connection ended.
2015-03-13 18:03:12,683 - ssmreceive - INFO - Restarting SSM.
2015-03-13 18:03:12,691 - ssm.ssm2 - INFO - Connecting using SSL...
2015-03-13 18:03:12,735 - stomp.py - WARNING - Could not connect to host mq.cro-ngi.hr, port 6162: (111, 'Connection refused')
2015-03-13 18:03:12,978 - stomp.py - WARNING - Could not connect to host mq.cro-ngi.hr, port 6162: (111, 'Connection refused')
2015-03-13 18:03:13,221 - stomp.py - WARNING - Could not connect to host mq.cro-ngi.hr, port 6162: (111, 'Connection refused')
2015-03-13 18:03:13,620 - ssm.ssm2 - WARNING - Failed to connect to mq.cro-ngi.hr:6162.
2015-03-13 18:03:13,620 - ssm.ssm2 - INFO - Connecting using SSL...
2015-03-13 18:03:13,664 - stomp.py - WARNING - Could not connect to host mq.cro-ngi.hr, port 6162: (111, 'Connection refused')
2015-03-13 18:03:13,908 - stomp.py - WARNING - Could not connect to host mq.cro-ngi.hr, port 6162: (111, 'Connection refused')
2015-03-13 18:03:14,152 - stomp.py - WARNING - Could not connect to host **mq.cro-ngi.hr**, port 6162: (111, 'Connection refused')
2015-03-13 18:03:14,552 - ssm.ssm2 - WARNING - Failed to connect to **mq.afroditi.hellasgrid.gr:6162**.
2015-03-13 18:03:14,552 - ssmreceive - ERROR - Unexpected exception: Attempts to start the SSM failed.  The system will exit.
2015-03-13 18:03:14,552 - ssmreceive - ERROR - Exception type: ssm.ssm2.Ssm2Exception
2015-03-13 18:03:14,552 - ssmreceive - ERROR - The SSM will exit.
[...]

PR #30, opened 2015-09-09
Add delay to receiver restart on lost connection

tofu-rocketry:
Workaround for issue #26, but a proper fix should be found for that.

This delays the attempt to reconnect for ten minutes by which time broker problems are generally resolved.


tofu-rocketry:
Two lines with no unit test coverage added, hence the fall in coverage, but they've worked fine on the dev machine.

stuartpullinger:
Thanks. OK to merge.

PR #32, opened 2015-11-26
Update expired cert and signed messages

tofu-rocketry:
Test cert has expired so needed updating and the messages needed re-signing.


tofu-rocketry:
Coverage increase seems to be from Coveralls correcting how it counts relevant lines.

PR #33, opened 2016-02-09
Ssm setup script

### LINKED ISSUE #31: Add setup.py script
We should add a `setup.py` script to allow our packages to be installed on any OS that has Python installed. It would also possibly help with automated testing.

https://docs.python.org/2.6/install/
https://docs.python.org/2.6/distutils/setupscript.html

Same issue for APEL: apel/apel#86

gregcorbett:
resolves #31 

a setup script for the SSM, usage: python setup.py install

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,0 +1,30 @@
+from setuptools import setup, find_packages
+
+# required to parse the requiremnts.txt file
+# from pip.req import parse_requirements
+
+# parse_requirements() returns generator of pip.req.InstallRequirement objects
+# install_requirements = parse_requirements("requirements.txt")
+# print str(install_reqs).strip('[]')
+
+# nned install_requirements as a list of requirements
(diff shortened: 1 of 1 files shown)


tofu-rocketry:
I've tested this out now. I've only spotted a couple of differences to what you end up with using the RPM:
- `setup.py` actually installs `logging.conf`. (This may be a bug for the RPM version.)
- The logrotate script has a different name. (Uses repo name of `ssm.logrotate` rather than RPM name of `apel-ssm`.) Seems about equivalent.
- init script ends up in wrong place - should be `/etc/init.d` rather than `bin`.

I've got a few commits to add. One of which fixes the last point.

tofu-rocketry:
QuatifiedCode warning is about use of string interpolation operator (`%`) rather than `format()`. Not too worried about that as rest of code base follows the former style.

PR #34, opened 2016-08-26
Update Python versions in Travis file

tofu-rocketry:
As per apel/apel#105.

- Remove Python 2.7 from allowed failures as we should now be building for
  OSes that have it as default.
- Update Python 3.4 to 3.5 as we should just check the latest available.


gregcorbett:
looks good

PR #35, opened 2016-10-06
Add python-devel as BuildRequires

alphacc:
The spec file assume python is available in the buildroot, in our case it is not by default.

According to fedora/rhel convention BuildRequires:  python-devel is needed.

If we are targeting also rhel5 python-devel otherwise it can be python2-devel


alphacc:
Hi,

We do not use fedora-packager in our buildroot for Centos7.

We follow EPEL recommendation fro packaging python :
https://fedoraproject.org/wiki/Packaging:Python

The RPM should be able to build anywhere and warn if there are missing BuildRequires (which is not the case for fedora-packager either.)

Thomas

tofu-rocketry:
`python-devel` is more than is required though, so could `BuildRequires:  python` just be used instead?

tofu-rocketry:
Merged in #47.

PR #38, opened 2017-04-03
Replace hard coded cert/key used in testing with generated ones

gregcorbett:
This means we wont have to regenerate the hard coded values once a year.

I have removed the hard coded values and replaced them with constants that should contain the key, cert and ca directory path. Keys, certs and ca files are then generated for each test.

most of the changes were simply 
* `self.certpath` and `self.keypath` to `TEST_CERT_FILE` and `TEST_KEY_FILE`
* `self.ca_dir` to `TEST_CA_DIR`

I also removed `test_from_file`, as there is no need to check the file contains the certificate it should, as the cert file is now dynamically generated

Encrypted or signed messages are generated in the test case if needed. they could be done in `setUp` but are only needed for a few test cases.

### DIFF
--- a/test/test_crypto.py
+++ b/test/test_crypto.py
@@ -51,7 +51,7 @@ class TestEncryptUtils(unittest.TestCase):
         with open(TEST_CERT_FILE, 'r') as test_cert:
             cert_string = test_cert.read()
 
-        hash_name, unused_error = p1.communicate(cert_string)
+        hash_name, _unused_error = p1.communicate(cert_string)
 
         self.ca_certpath = os.path.join(TEST_CA_DIR, hash_name.strip() + '.0')
         with open(self.ca_certpath, 'w') as ca_cert:
--- a/test/test_crypto.py
+++ b/test/test_crypto.py
@@ -8,7 +8,7 @@ import logging
 import os
 import tempfile
 import shutil
-from subprocess import Popen, PIPE
+from subprocess import call, Popen, PIPE
 
 from ssm.crypto import _from_file, \
     check_cert_key, \
@@ -33,43 +33,35 @@ class TestEncryptUtils(unittest.TestCase):
     
     def setUp(self):
         '''
-        Some of the functions require the certificate and key to be files.
+        If no key/cert pair found, generate a new
+        key/cert pair and store as a file.
         '''
-        self.certfile, self.certpath = tempfile.mkstemp(prefix='cert')
-        os.write(self.certfile, TEST_CERT)
-        os.close(self.certfile)
+        # create a key/cert pair
+        call(['openssl', 'req', '-x509', '-nodes', '-days', '1',
+              '-newkey', 'rsa:2048', '-keyout', TEST_KEY_FILE,
+              '-out', TEST_CERT_FILE, '-subj', TEST_CERT_DN])
 
-        self.keyfile, self.keypath = tempfile.mkstemp(prefix='key')
-        os.write(self.keyfile, TEST_KEY)
-        os.close(self.keyfile)
-        
         # Set up an openssl-style CA directory, containing the 
         # self-signed certificate as its own CA certificate, but with its
         # name as <hash-of-subject-DN>.0.
-        
-        ca_certs = [TEST_CERT]
+        p1 = Popen(['openssl', 'x509', '-subject_hash', '-noout'],
+                    stdin=PIPE, stdout=PIPE, stderr=PIPE)
 
-        self.ca_dir = tempfile.mkdtemp(prefix='ca')
-        for cert in ca_certs:
-            p1 = Popen(['openssl', 'x509', '-subject_hash', '-noout'],
-                       stdin=PIPE, stdout=PIPE, stderr=PIPE)
-            hash_name, unused_error = p1.communicate(cert)
-            ca_certpath = os.path.join(self.ca_dir, hash_name.strip() + '.0')
-            ca_cert = open(ca_certpath, 'w')
-            ca_cert.write(cert)
-            ca_cert.close()
+        with open(TEST_CERT_FILE, 'r') as test_cert:
+            cert_string = test_cert.read()
+
+        hash_name, unused_error = p1.communicate(cert_string)
+
+        self.ca_certpath = os.path.join(TEST_CA_DIR, hash_name.strip() + '.0')
+        with open(self.ca_certpath, 'w') as ca_cert:
+            ca_cert.write(cert_string)
 
     def tearDown(self):
-        '''
-        Remove temporary files.
-        '''
-        os.remove(self.certpath)
-        os.remove(self.keypath)
-        
-        # Remove the CA dir and any contents.
-        shutil.rmtree(self.ca_dir)
-        
-    
+        '''Remove temporary files.'''
+        os.remove(TEST_CERT_FILE)
+        os.remove(TEST_KEY_FILE)
+        os.remove(self.ca_certpath)
+ 
     def test_from_file(self):
         '''
         Just test that the temporary file that has been set up contains the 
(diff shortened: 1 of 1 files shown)


test/test_crypto.py
```diff
@@ -33,53 +33,36 @@ class TestEncryptUtils(unittest.TestCase):
...
-        Some of the functions require the certificate and key to be files.
+        If no key/cert pair found, generate a new
+        key/cert pair and store as a file.
         '''
-        self.certfile, self.certpath = tempfile.mkstemp(prefix='cert')
-        os.write(self.certfile, TEST_CERT)
-        os.close(self.certfile)
+        if not (os.path.exists(TEST_CERT_FILE) and os.path.exists(TEST_KEY_FILE)):
+            # create a key/cert pair
+            call(['openssl', 'req', '-x509', '-nodes', '-days', '365',
```
tofu-rocketry: As we're generating certs as we test, the expiry should probably be much shorter than a year.

test/test_crypto.py
```diff
@@ -33,53 +33,36 @@ class TestEncryptUtils(unittest.TestCase):
...
-            ca_cert = open(ca_certpath, 'w')
-            ca_cert.write(cert)
-            ca_cert.close()
+
+            with file(TEST_CERT_FILE) as test_cert:
+                cert_string = test_cert.read()
+                hash_name, unused_error = p1.communicate(cert_string)
+
+                self.ca_certpath = os.path.join(TEST_CA_DIR, hash_name.strip() + '.0')
+                ca_cert = open(self.ca_certpath, 'w')
```
tofu-rocketry: Guess you could use another `with` here.

test/test_crypto.py
```diff
@@ -33,53 +33,36 @@ class TestEncryptUtils(unittest.TestCase):
...
             p1 = Popen(['openssl', 'x509', '-subject_hash', '-noout'],
                        stdin=PIPE, stdout=PIPE, stderr=PIPE)
-            hash_name, unused_error = p1.communicate(cert)
-            ca_certpath = os.path.join(self.ca_dir, hash_name.strip() + '.0')
-            ca_cert = open(ca_certpath, 'w')
-            ca_cert.write(cert)
-            ca_cert.close()
+
+            with file(TEST_CERT_FILE) as test_cert:
+                cert_string = test_cert.read()
```
tofu-rocketry: You've finished using `test_cert` by this stage, so should dedent the following lines.

test/test_crypto.py
```diff
@@ -33,53 +33,36 @@ class TestEncryptUtils(unittest.TestCase):
...
+            # name as <hash-of-subject-DN>.0.
             p1 = Popen(['openssl', 'x509', '-subject_hash', '-noout'],
                        stdin=PIPE, stdout=PIPE, stderr=PIPE)
-            hash_name, unused_error = p1.communicate(cert)
-            ca_certpath = os.path.join(self.ca_dir, hash_name.strip() + '.0')
-            ca_cert = open(ca_certpath, 'w')
-            ca_cert.write(cert)
-            ca_cert.close()
+
+            with file(TEST_CERT_FILE) as test_cert:
```
tofu-rocketry: [`open()` is recommended over `file()`.](https://docs.python.org/2/library/functions.html#file) Also the latter's not available in Python 3.

There's a few other occurrences.

test/test_crypto.py
```diff
@@ -33,53 +33,36 @@ class TestEncryptUtils(unittest.TestCase):
...
-        Remove temporary files.
-        '''
-        os.remove(self.certpath)
-        os.remove(self.keypath)
-        
-        # Remove the CA dir and any contents.
-        shutil.rmtree(self.ca_dir)
-        
-    
-    def test_from_file(self):
```
tofu-rocketry: I take it that this test is superfluous now?

test/test_crypto.py
```diff
@@ -192,32 +180,37 @@ def test_get_signer_cert(self):
...
-        
-        if cert.strip() != TEST_CERT:
-            self.fail('Certificate retrieved from signature does not match \
-                    certificate used to sign.')
+
+        with file(TEST_CERT_FILE) as test_cert:
+            cert_string = test_cert.read()
+
+        if cert.strip() != cert_string.strip():
+            self.fail(('Certificate retrieved from signature'
```
tofu-rocketry: You shouldn't need the extra parentheses to do implied line continuation.

test/test_crypto.py
tofu-rocketry: Oh, yes. You mention that in the PR!

test/test_crypto.py
```diff
@@ -33,53 +33,36 @@ class TestEncryptUtils(unittest.TestCase):
...
     def setUp(self):
         '''
-        Some of the functions require the certificate and key to be files.
+        If no key/cert pair found, generate a new
+        key/cert pair and store as a file.
         '''
-        self.certfile, self.certpath = tempfile.mkstemp(prefix='cert')
-        os.write(self.certfile, TEST_CERT)
-        os.close(self.certfile)
+        if not (os.path.exists(TEST_CERT_FILE) and os.path.exists(TEST_KEY_FILE)):
```
gregcorbett: yeah there is no need for the if, removed

test/test_crypto.py
gregcorbett: reduced to a day, that seems to be the shortest we can do with the `openssl` command line tool

test/test_crypto.py
gregcorbett: `file()` changed to `open()`, throughout the file

test/test_crypto.py
gregcorbett: dedented

test/test_crypto.py
gregcorbett: changed

test/test_crypto.py
gregcorbett: removed

gregcorbett:
changes made, I'll track down the 2 lines of lost coverage - but I'm having trouble viewing the coveralls reports (I'm hoping that changes when they are regenerated!)

tofu-rocketry:
I've just changed the pathing in Coveralls and the source is showing again. Pretty sure I did try that specific file path before though! 😕

test/test_crypto.py
tofu-rocketry: Yeah, that's the shortest I think you can do without diving into madness/low-level options.

test/test_crypto.py
tofu-rocketry: It's not this test that was somehow hitting those extra lines, surely? 😟

gregcorbett: it looks like its lines [168](https://github.com/apel/ssm/blob/dev/ssm/crypto.py#L168) and [179](https://github.com/apel/ssm/blob/dev/ssm/crypto.py#L179) that were being hit but now aren't. I'll look into why

test/test_crypto.py
```diff
@@ -33,53 +33,35 @@ class TestEncryptUtils(unittest.TestCase):
...
-        self.ca_dir = tempfile.mkdtemp(prefix='ca')
-        for cert in ca_certs:
-            p1 = Popen(['openssl', 'x509', '-subject_hash', '-noout'],
-                       stdin=PIPE, stdout=PIPE, stderr=PIPE)
-            hash_name, unused_error = p1.communicate(cert)
-            ca_certpath = os.path.join(self.ca_dir, hash_name.strip() + '.0')
-            ca_cert = open(ca_certpath, 'w')
-            ca_cert.write(cert)
-            ca_cert.close()
+        hash_name, unused_error = p1.communicate(cert_string)
```
tofu-rocketry: Pylint complains about unused variables. A leading underscore (i.e. `_unused_error`) should suppress it and make it clear that it shouldn't be used. I think that's a bit nicer than indexing the right hand side in this case.

test/test_crypto.py
tofu-rocketry: At least Coveralls now finally highlights the lines that have changed coverage with icons. 😀

gregcorbett: ah yes, it's definitely those lines then

gregcorbett:
it looks like the two lines difference is because `SIGNED_MSG2` previously had it's 

> \# ... text manually converted to quoted-printable
> \# and Content-Type and Content-Transfer-Encoding fields added in manually.

Where as now it just generated by the `crypto.sign` method. I'll look into automating that step

test/test_crypto.py
gregcorbett: I still get the pylint error with `_unused_error` 

`W: 54,19: Unused variable '_unused_error' (unused-variable)`

By indexing, do you mean using `_` instead of a named variable?

gregcorbett: although I now get what you mean by 'make it clear that it shouldn't be used'

tofu-rocketry: Bah. I've figured it out. Newest versions of PyLint should ignore variables with leading underscore, but Codacy is tied to an older version*. So, keep it as a single underscore, and we'll just ignore the error. (And it may even sort itself out in the future!)

*It uses a PyLint plugin written by the Landscape.io dev in fact, that limits the PyLint version to less than 1.4 which is the version of Pylint that dropped support for Python 2.6.

tofu-rocketry: By indexing I mean using something like `p1.communicate(cert_string)[0]` to get just the value we want.

gregcorbett:
changes made

tofu-rocketry:
### Codacy failure
Newest versions of PyLint should ignore variables with leading underscore, but Codacy is tied to an older version*. So we'll just ignore the error. (And it may even sort itself out in the future!)

### Coverage failure
Change to the way `test_verify` works exposed an oddity in the way `crypto.verify` works. As this PR is just to remove the need to re-generate certs regularly, will ignore for now, but will look at problem in issue #39.
___
*It uses a PyLint plugin written by the Landscape.io dev in fact, that [limits the PyLint version to less than 1.4](https://github.com/landscapeio/pylint-common/blob/master/setup.py#L30) which is the version of Pylint that dropped support for Python 2.6.

PR #40, opened 2017-04-24
Prevent an SSM from being instantiated with an expired certificate

gregcorbett:
Currently, an SSM can be instantiated with an expired certificate. The APEL server then rejects the messages sent with that SSM due to the signature being invalid.

This PR prevents an SSM from being created with an expired certificate, preventing messages that would otherwise be rejected from being sent to the server. 

It does this by:
- Adding a `verify_cert_date` to `crypto` which returns True if the certificate has not yet expired, otherwise it returns False.
- Making `ssm.__init__()` raise an Ssm2Exception if `crypto.verify_cert_date()` detects an expired certificate is being used.

It also reworks `test/test_ssm.py` to:
- Generate a new certificate on the fly, using a hardcoded key
- Create a new SSM for each test case, so that some tests can have the SSM creation fail
- Adds a test checking an SSM cannot be created using an expired certificate

The expired certificate and corresponding key are hardcoded as we can't generate a certificate that expires in the time frame of a unit test with the openssl command line tool.


gregcorbett:
Looks good, I have rebased

PR #42, opened 2017-04-25
Improvements to setup.py script

gregcorbett:
- `python setup.py install` now installs all files the RPM would
- `python setup.py test` can now run the test suite (without needing to alter the python path)
- Improve commenting of options like `zip_safe`


gregcorbett:
I've made `test/__init__.py` blank and disabled `zip_safe`, as discussed yesterday.

gregcorbett:
rebased

PR #44, opened 2017-05-03
Change Travis file to test against Python nightly

tofu-rocketry:
Same as for apel/apel#134.

Change Travis file to test against latest nightly rather than a fixed Python 3 version so that it doesn't have to continually be manually changed.

At the moment this will be Python 3.7.0a0.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,10 +2,10 @@ language: python
 python:
   - "2.6"
   - "2.7"
-  - "3.5"
+  - "nightly"
 matrix:
   allow_failures:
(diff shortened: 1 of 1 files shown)


tofu-rocketry:
Ah, this will need to rebased on the crypto fix at some point.

tofu-rocketry:
Sorry. I had to rebase and I've got it set to dismiss reviews on a new push.

tofu-rocketry:
I'd be quite worried if it didn't. 😛

PR #47, opened 2017-09-13
Alphacc patch 1

tofu-rocketry:
Closes #35.

Rebased version of #35 with additional comment. As this change doesn't affect our builds due to the conditional, I'm okay with the build requirement of python-devel, even though this seems to be over specifying the package required (only python is actually needed to build as this isn't a Python extension - which would require python-devel).


tofu-rocketry:
Coverage decrease is due to branch coverage being enabled since presumably the last time `dev` was build in Travis.

PR #51, opened 2017-11-13
Merge two log messages into one and change level

tofu-rocketry:
- Merge refresh and ping log messages into one as don't really need to
have two when they always occur together.
- Reduce logging level of connection lost to warning as software is
still working correctly.

### DIFF
--- a/bin/receiver.py
+++ b/bin/receiver.py
@@ -183,15 +183,14 @@ def main():
             time.sleep(1)
 
             if i % REFRESH_DNS == 0:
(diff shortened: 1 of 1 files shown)


gregcorbett:
This will have an effect on the ELK monitoring. `Refreshing the valid DNs` is one of the lines used by filebeat to separate multi line events. 

But, it would be a minor change of the filebeat config on our machines, which can be done while the upgrade is taking place, so it isn't a blocking issue.

PR #55, opened 2018-04-25
Correct files that have CRLF line endings to LF

### LINKED ISSUE #53: Expunge all CRLF line endings
This seems to be the only offending file:

```
$ grep -rUIl $'\r' *
bin/__init__.py
```

tofu-rocketry:
Resolves #53.


tofu-rocketry:
And with that, all the CRLFs are gone! Let's avoid reintroducing them. :imp:

PR #56, opened 2018-04-30
Stop setup.py installing /etc/apel/logging.cfg

gregcorbett:
As it is not installed by the RPM and having it installed causes confusion when the logging config section is only specified in the sender or receiver config but `logging.cfg` is under `/etc/apel`, as in that case it takes precedence over the supplied configs logging section.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -39,8 +39,7 @@ def main():
 
     # conf_files will later be copied to conf_dir
     conf_dir = '/etc/apel/'
-    conf_files = ['conf/logging.cfg',
-                  'conf/receiver.cfg',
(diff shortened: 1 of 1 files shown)


tofu-rocketry:
One question, though not directly related to this: does `setup.py` install the logrotate script with the same name that's used by the RPM?

gregcorbett:
I believe it does, `setup.py` installs the logrotate script under `/etc/logrotate.d/` with the file name `apel-ssm`

tofu-rocketry:
Ok, I have a note on the inconsistencies which is out of date then.

PR #58, opened 2018-05-01
Add comment to logging.cfg to clarify usage

### LINKED ISSUE #57: Add explanatory comments to logging.cfg
See #56.

Should add some comments explaining it's for advanced use only and not installed by RPM/setup.py.

tofu-rocketry:
Resolves #57.

`logging.cfg` seems to have only ever caused problems. So some explanatory comments should help avoid problems.

### DIFF
--- a/conf/logging.cfg
+++ b/conf/logging.cfg
@@ -1,3 +1,7 @@
+# This logging configuration file is not installed by RPM or setup.py, but if
+# placed in /etc/apel alongside the other config files it will override them.
+# Advanced use only. See the Python docs for the "logging.config" module.
+
 [loggers]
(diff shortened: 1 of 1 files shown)


jrha:
So how *would* it get distributed?

tofu-rocketry:
Well this file doesn't get distributed, but it should be kept as a template as there is functionality built into the Python code to support it if it is present.

jrha:
I guess my question is really: what's the point in the functionality if you don't want it being used? (The original issue seems to consider this functionality harmful).

tofu-rocketry:
Removing the functionality should be possible, but would require careful deprecation in case someone is using it.

PR #59, opened 2018-05-02
Expired server cert check

tofu-rocketry:
Follow on from #40. This is fairly minor and in a sense fixes an issue for clients so I'm including it in 2.2.1.

This PR:
- changes the comment in the config for `server_cert` to clarify its usage
- adds checking of the server cert expiry
- adds corresponding unit test
- tidies a little whitespace.

### DIFF
--- a/conf/sender.cfg
+++ b/conf/sender.cfg
@@ -24,8 +24,10 @@ use_ssl: true
 certificate: /etc/grid-security/hostcert.pem
 key: /etc/grid-security/hostkey.pem
 capath: /etc/grid-security/certificates
+
 # If supplied, outgoing messages will be encrypted using this certificate.
-# It must be the cert of the accounting repository that you are sending to.
+# May be used in addition to 'use_ssl'. If used, it must be the certificate of
+# the final server that's receiving your messages; not your own, nor the broker.
 #server_cert: /etc/grid-security/servercert.pem
 
 ################################################################################
--- a/conf/sender.cfg
+++ b/conf/sender.cfg
@@ -24,8 +24,8 @@ use_ssl: true
 certificate: /etc/grid-security/hostcert.pem
 key: /etc/grid-security/hostkey.pem
 capath: /etc/grid-security/certificates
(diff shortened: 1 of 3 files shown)


ssm/ssm2.py
```diff
@@ -112,6 +112,12 @@ def __init__(self, hosts_and_ports, qpath, cert, key, dest=None, listen=None,
             log.info('Messages will be encrypted using %s', enc_cert)
             if not os.path.isfile(self._enc_cert):
                 raise Ssm2Exception('Specified certificate file does not exist: %s.' % self._enc_cert)
+            # Check that the encyption certificate has not expired.
+            if not crypto.verify_cert_date(enc_cert):
+                raise Ssm2Exception(
+                    'Encryption certificate %s has expired. Please obtain the '
+                    'current one for the server you are sending to.' % enc_cert
```
gregcorbett: Maybe change this to `Please obtain the current one for the accounting repository you are sending to`?

Don't want people confusing the 'server' for the broker

conf/sender.cfg
```diff
@@ -24,8 +24,8 @@ use_ssl: true
 certificate: /etc/grid-security/hostcert.pem
 key: /etc/grid-security/hostkey.pem
 capath: /etc/grid-security/certificates
-# If this is supplied, outgoing messages will be encrypted
-# using this certificate
+# If supplied, outgoing messages will be encrypted using this certificate.
+# It must be the cert of the accounting repository that you are sending to.
```
gregcorbett: Maybe change `cert` to `certificate`?

gregcorbett: and maybe we could mention that this form of encryption is above and beyond encrypting the tunnel with SSL?

tofu-rocketry: Addressed, though I've also made it less accounting specific.

ssm/ssm2.py
tofu-rocketry: Changed somewhat, though I've also made it less accounting specific.

PR #72, opened 2018-06-15
Support for non-dirq sending

### LINKED ISSUE #50: Add support for non-dirq sending
To enable new sources of accounting data to be added more easily (e.g. new storage systems like Echo), it would be good to allow the sending of messages from directories that are not structured as required by the dirq module.

Assuming a site is running a sending SSM on a cron after some kind of accounting data extraction process, we don't really need all the fancy FIFO and other functionality that dirq provides, and structuring the output of their extraction process (in the specific way that is required by dirq) seems an unnecessary extra step for a 3rd party script developer.

We should only require those devs to write files that match one of our accepted record formats and not burden them with faffing with directory structure. Just put the files together in one directory that we can point SSM to.

I imagine that:
- we will require a new config option to say that a non-dirq format is being used (default it to dirq format for backwards compatibility).
- when we create the list of files to send after inspecting the directory, we might want to at last sort them by creation time to give some kind of FIFOness.
- this feature should be limited to use by sending SSMs in cron mode (i.e. non-daemons).

The "Simple™️"  way seems to be to replicate the dirq interface for the dirq methods used in `send_all` and `has_msgs` and then just choose which "directory queue" object is used when creating the `Ssm2` sender instance. Is there a more elegant solution?

gregcorbett:
resolves #50 

It does this by replicating the dirq interface for the dirq methods used in `send_all` and `has_msgs` and then allows the user to choose which "directory queue" object is used when creating the Ssm2 sender instance.

It also:

- adds a new config option (`path_type`) to say that a non-dirq format is being used. It's commented out and hence uses the dirq format by default for backwards compatibility.
- sends files in order they were last modified, to give some kind of FIFOness.
- has exception handling for when files are not readable.
- limits this functionality to sending SSMs
- removes `python-dirq` as a requirement in the spec file and lists it as a extra requirement in the setup.py file.
- updates documentation about how to add files for sending


conf/sender.cfg
```diff
@@ -40,6 +40,10 @@ destination:
 
 # Outgoing messages will be read and removed from this directory.
 path: /var/spool/apel/outgoing
+# If 'path_type: directory' is supplied, the supplied 'path' will be
+# treated as if it is a directory rather than a dirq. As a result,
+# 'path' cannot contain subdirectories.
+#path_type: directory
```
tofu-rocketry: This should be uncommented but default to 'dirq'. Existing users won't get their config overwritten. The description will need updating to clarify both options.

ssm/message_directory.py
```diff
@@ -0,0 +1,138 @@
...
+class MessageDirectory(object):
+    """A structure for holding Accounting messages in a directory."""
+
+    def __init__(self, path):
+        """Create a new directory structure for holding Accounting messages."""
+        self.directory_path = path
+
+    def add(self, data):
+        """Add the passed data to a new file and return it's name."""
+        # Use uuid4 to create a file name because uuid4 is
```
tofu-rocketry: This comment reads oddly. Really, we want to know why we're creating a unique name.

ssm/message_directory.py
```diff
@@ -0,0 +1,138 @@
...
+
+@author: Greg Corbett
+"""
+
+import logging
+import os
+import uuid
+
+# logging configuration
+LOG = logging.getLogger(__name__)
```
tofu-rocketry: `LOG` should not be capitalised. It is not a constant.

ssm/message_directory.py
```diff
@@ -0,0 +1,138 @@
...
+        """Return True to simulate a successful lock. Does nothing else."""
+        return True
+
+    def purge(self):
+        """
+        Do nothing, as there are no old/intermediate directories to purge.
+
+        Only included to preserve dirq interface.
+        """
+        LOG.debug("purge() called, but purge() does nothing.")
```
tofu-rocketry: Should append something like "...for non-dirq sending."

ssm/ssm2.py
```diff
@@ -91,10 +99,29 @@ def __init__(self, hosts_and_ports, qpath, cert, key, dest=None, listen=None,
...
+
+            elif path_type == 'directory':
+                self._outq = MessageDirectory(qpath)
+            else:
+                raise Ssm2Exception('Unsupported path_type variable.')
+
         elif listen is not None:
             inqpath = os.path.join(qpath, 'incoming')
             rejectqpath = os.path.join(qpath, 'reject')
+
```
tofu-rocketry: Should check that `path_type == 'dirq'` and raise a warning if it's not and say something like the setting is wrong but is being ignored.

test/test_message_directory.py
```diff
@@ -0,0 +1,138 @@
+"""
```
tofu-rocketry: Don't put creation/authorship info into files that are controlled by a VCS. Leave that to the VCS.

conf/sender.cfg
gregcorbett: changed

ssm/message_directory.py
gregcorbett: changed

ssm/message_directory.py
gregcorbett: changed

ssm/message_directory.py
gregcorbett: changed

ssm/ssm2.py
gregcorbett: This check seems like it would be a good idea.

`bin/receiver.py` doesn't pass `path_type` through to the SSM at the moment. So I could make it pass it through and put the check in `ssm/ssm2.py` or make the check in `bin/receiver.py`, thoughts?

ssm/message_directory.py
```diff
@@ -0,0 +1,138 @@
...
+        # Use uuid4 to create a file name because uuid4 is
+        # sufficient to create an unique ID.
+        name = uuid.uuid4()
+
+        try:
+            # Open the file and write the provided data into the file.
+            with open("%s/%s" % (self.directory_path, name), 'w') as message:
+                message.write(data)
+        except (IOError, OSError) as error:
+            LOG.error("Could not create file %s/%s: %s",
```
gregcorbett: looking at this with fresh eyes, just logging the error seems like a bad idea. I don't think the add method should return a filename if a file was not created.

Thoughts @tofu-rocketry?

I can't quite parse what the dirq module would do in this case, see [here](https://github.com/cern-mig/python-dirq/blob/75f63e193c209edd40a7a643386100d953bcbdab/dirq/QueueSimple.py#L212) and [here](https://github.com/cern-mig/python-dirq/blob/75f63e193c209edd40a7a643386100d953bcbdab/dirq/QueueSimple.py#L178). Does it raise an OSError?

test/test_message_directory.py
gregcorbett: Should I copy the doc string from `ssm/message_directory.py` here (minus the creation/authorship info)?

ssm/message_directory.py
```diff
@@ -0,0 +1,138 @@
...
+
+        Only included to preserve dirq interface.
+        """
+        LOG.debug("purge() called, but purge() does nothing.")
+
+    def remove(self, name):
+        """Remove the named message."""
+        try:
+            os.unlink("%s/%s" % (self.directory_path, name))
+        except (IOError, OSError) as error:
```
gregcorbett: I'm having second thoughts about this error handling as well, logging the error but carrying on regardless. 

[dirq seems like it would let the exception raise up](https://github.com/cern-mig/python-dirq/blob/75f63e193c209edd40a7a643386100d953bcbdab/dirq/QueueSimple.py#L311)

test/test_message_directory.py
tofu-rocketry: I'd actually use the type of header used here (i.e. inline comment syntax): https://github.com/apel/apel/blob/dev/test/test_norm_sum_record.py

But I would drop the "The" from our name (and use 2018 of course).

Putting it in the file's docstring means it can pop up on the command line and via other tools in ways we probably don't intend it to.

conf/sender.cfg
tofu-rocketry: Also this should be first option in the messaging section.

ssm/message_directory.py
```diff
@@ -0,0 +1,138 @@
...
+
+
+class MessageDirectory(object):
+    """A structure for holding Accounting messages in a directory."""
+
+    def __init__(self, path):
+        """Create a new directory structure for holding Accounting messages."""
+        self.directory_path = path
+
+    def add(self, data):
```
tofu-rocketry: We don't actually use `add` for the outgoing queue. Could drop it?

ssm/message_directory.py
```diff
@@ -0,0 +1,138 @@
...
+            # If we get here, just return the arbitrarily ordered list.
+            return file_name_list
+
+        except (IOError, OSError) as error:
+            LOG.error(error)
+            # Return an empty file list.
+            return []
+
+
+    def __iter__(self):
```
tofu-rocketry: Is this used by the sender?

tofu-rocketry: Ah, right, it would be usef by `for msgid in self._outq`.

ssm/message_directory.py
```diff
@@ -0,0 +1,138 @@
...
+        # to keep the dirq like interface.
+        return "%s" % name
+
+    def count(self):
+        """
+        Return the number of elements in the queue.
+
+        Regardless of their state.
+        """
+        return len(self._get_messages())
```
tofu-rocketry: This is not very efficient as `_get_messages` does a lot of work. Is there a lighter way to do this?

ssm/ssm2.py
tofu-rocketry: You don't need to do any of that. `path_type` has `dirq` as the default in the `__init__` of `Ssm2`.

ssm/message_directory.py
tofu-rocketry: Well, we could have a stub class that raises a useful error, just in case.

conf/sender.cfg
gregcorbett: first option in the messaging section overall, or just before `path`?

ssm/ssm2.py
gregcorbett: but the default `path_type` currently doesn't get changed by `bin/receiver.py` - even if it's set it in `conf/receiver.cfg`,

test/test_message_directory.py
gregcorbett: changed

ssm/message_directory.py
gregcorbett: We don't use `dirq.add` at the moment. But I find the `dirq.add` functionality quite useful for running manual tests. And maybe in the future we could automate some of those tests and make use of this `add` method

ssm/message_directory.py
gregcorbett: 👍

ssm/message_directory.py
gregcorbett: without calling `_get_messages(sort_by_mtime=True)`, this is just a call to `os.listdir()` wrapped in a `try...except`

ssm/message_directory.py
```diff
@@ -0,0 +1,138 @@
...
+                    # Store the sorted file paths in a class element.
+                    sorted_messages.append(file_name)
+
+                # Return the sorted list.
+                return sorted_messages
+
+            # If we get here, just return the arbitrarily ordered list.
+            return file_name_list
+
+        except (IOError, OSError) as error:
```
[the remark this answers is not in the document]
gregcorbett: hmmm maybe? The reason for catching it was to show the user a nicer log line, rather than a stack trace

gregcorbett:
There is still this comment that GitHub has hidden: https://github.com/apel/ssm/pull/72#discussion_r216679788

ssm/ssm2.py
tofu-rocketry: Ok, we'll not worry about this then.

ssm/message_directory.py
tofu-rocketry: Fair enough. Was just thinking of avoiding complexity.

ssm/message_directory.py
tofu-rocketry: You should either just use `listdir` directly, or make the sorting part of `_get_messages` optional.

tofu-rocketry: Oh, I misread - it is already. Make the call explicit here then.

ssm/message_directory.py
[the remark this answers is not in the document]
tofu-rocketry: Ok.

ssm/message_directory.py
```diff
@@ -0,0 +1,132 @@
...
+        """
+        Do nothing, as there are no old/intermediate directories to purge.
+
+        Only included to preserve dirq interface.
+        """
+        log.debug("purge called, but purge does nothing for non-dirq sending.")
+
+    def remove(self, name):
+        """Remove the named message."""
+        os.unlink("%s/%s" % (self.directory_path, name))
```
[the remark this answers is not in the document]
gregcorbett: I used `unlink` because that's what the `dirq.remove` method uses.

Do you mean `os.remove` is used, else where in ssm? or else where in this file?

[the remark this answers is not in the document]
tofu-rocketry: Elsewhere in SSM.

ssm/message_directory.py
```diff
@@ -0,0 +1,132 @@
...
+class MessageDirectory(object):
+    """A structure for holding Accounting messages in a directory."""
+
+    def __init__(self, path):
+        """Create a new directory structure for holding Accounting messages."""
+        self.directory_path = path
+
+    def add(self, data):
+        """Add the passed data to a new file and return it's name."""
+        # Create a unique file name so APEL admins can pair sent and recieved
```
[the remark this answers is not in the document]
gregcorbett: changed

ssm/message_directory.py
```diff
@@ -0,0 +1,132 @@
...
+
+# logging configuration
+log = logging.getLogger(__name__)
+
+
+class MessageDirectory(object):
+    """A structure for holding Accounting messages in a directory."""
+
+    def __init__(self, path):
+        """Create a new directory structure for holding Accounting messages."""
```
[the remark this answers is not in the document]
gregcorbett: I've dropped the reference to accounting, I think renaming the class to PlainQueue will also help with this - as directory does currently get used in the two contexts

README.md
```diff
@@ -93,26 +94,38 @@ configuration will send messages to the test apel server.
...
 
-There are two ways to add files to be sent:
+There are multiple manual and programmatic ways to add files to be sent: 
 
 ### Manual
 
+#### With the dirq module
 All file and directory names must use hex characters: `[0-9a-f]`.
 
  * Create a directory within /var/spool/apel/outgoing with a name
```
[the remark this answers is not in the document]
gregcorbett: Do you mean like this

> #### With the `dirq` module

?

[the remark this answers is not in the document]
tofu-rocketry: I just meant the file paths, as that style is used later on.

ssm/message_directory.py
```diff
@@ -0,0 +1,132 @@
+# Copyright 2018 Science and Technology Facilities Council
```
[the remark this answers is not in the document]
gregcorbett: done

ssm/message_directory.py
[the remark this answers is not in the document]
gregcorbett: reworded as part of renaming the class/file

ssm/message_directory.py
tofu-rocketry: dirq loops until it can create the directory or an error other than it already exists is raised.

Prob just raise the error.

tofu-rocketry:
Rebased (fixing a couple of conflicts) and [made some tweaks](https://github.com/apel/ssm/pull/72/files/4f9c03dccb9f41ab0654368d25ba205af9f90f02..6899574895e4df89100de44c64ee379611ee8452).

README.md
```diff
@@ -119,26 +120,38 @@ configuration will send messages to the test apel server.
 
 ## Adding Files
 
-There are two ways to add files to be sent:
+There are multiple manual and programmatic ways to add files to be sent:
```
jrha: white-space at EOL

bin/sender.py
```diff
@@ -120,9 +120,18 @@ def main():
                 raise Ssm2Exception('No destination queue is configured.')
         except ConfigParser.NoOptionError, e:
             raise Ssm2Exception(e)
-    
+
+        # Determine what type of message store we are interacting with,
+        # i.e. a dirq QueueSimple object or a plain MessageDirectory directory.
+        try:
+            path_type = cp.get('messaging', 'path_type')
```
jrha: What happens if the "messaging" section isn't defined?

ssm/message_directory.py
```diff
@@ -0,0 +1,127 @@
...
+    def _get_messages(self, sort_by_mtime=False):
+        """
+        Get the messages stored in this MessageDirectory.
+
+        if sort_by_mtime is set to True, the returned list is guaranteed to be
+        in increasing order of modification time.
+
+        mtime is used because (apparently) there is not way to find the
+        original date of file creation due to a limitation
+        of the underlying filesystem
```
jrha: It's a POSIX limitation (see https://lwn.net/Articles/397442/), Windows does provide a creation time thouh, but then allows it to be modified, which probably makes it useless.

If you create the messages though, why not encode the creation time in their filename?

README.md
```diff
@@ -119,26 +120,38 @@ configuration will send messages to the test apel server.
...
 
 ### Manual
 
+#### With the dirq module
 All file and directory names must use hex characters: `[0-9a-f]`.
 
  * Create a directory within /var/spool/apel/outgoing with a name
    of EIGHT hex characters e.g. `12345678`
  * Put files in this directory with names of FOURTEEN hex 
    e.g. `1234567890abcd`
```
jrha: I know you didn't touch this, but do the first eight characters have to match those in the directory name?

tofu-rocketry: No, they're independent: https://dirq.readthedocs.io/en/latest/queue.html#directory-structure

jrha: That's what I thought, but then the example uses the same characters...

ssm/message_directory.py
tofu-rocketry: We don't want any particular format of name to be a requirement as the point of this change is to allow the option of arbitrary files to be sent by SSM without using the `dirq` module and it's finicky naming scheme. We thought that sorting files by age before sending would at least give it some rough FIFO-ness.

bin/sender.py
tofu-rocketry: See lines below - we default to using `dirq` (which is the existing type).

tofu-rocketry:
Could you have a look at [my changes](https://github.com/apel/ssm/pull/72/files/4f9c03dccb9f41ab0654368d25ba205af9f90f02..4fd0d287d9a6ae599e764ddc343fa6155aa3b8ea) @gregcorbett ?

gregcorbett:
@tofu-rocketry, [your changes](https://github.com/apel/ssm/pull/72/files/4f9c03dccb9f41ab0654368d25ba205af9f90f02..4fd0d287d9a6ae599e764ddc343fa6155aa3b8ea) look good to me!

tofu-rocketry:
Any final thoughts @jrha ?

PR #81, opened 2018-09-20
Split test requirements out and add constraint

tofu-rocketry:
The weekly cron build failed which raised an issue with a dependency of a dependency not supporting Python 2.6 any more. This PR fixes that (though the longer term fix is to deprecate Python 2.6 support).

- Move test requirements into separate requirements file to make it
clearer that they are only needed for running unit tests.
- Add constraints file called from the test requirements file to limit
one of coveralls' dependencies to an earlier version that works with
Python 2.6.

### DIFF
--- a/requirements-test.txt
+++ b/requirements-test.txt
@@ -1,0 +1,6 @@
+# Constraints on requirements below
+-c constraints.txt
+# Requirements for testing
+unittest2
+coveralls
+mock
--- a/requirements-test.txt
+++ b/requirements-test.txt
@@ -1,6 +1,8 @@
-# Constraints on requirements below
+# Additional requirements for the unit and coverage tests
+
+# Constraints on the requirements below
 -c constraints.txt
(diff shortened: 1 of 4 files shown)


requirements-test.txt
```diff
@@ -0,0 +1,6 @@
+# Constraints on requirements below
+-c constraints.txt
```
gregcorbett: Does this work due to the way `-r` handles `requirements-test.txt`?
Could we have `pip install -r requirements-test.txt -c constraints.txt` in the install section of .travis.yml instead?

tofu-rocketry: See https://github.com/apel/apel/pull/171#discussion_r220097100.

gregcorbett: fair enough

gregcorbett:
I am happy to approve the changes, do you want to squash the commits down before I review though?

PR #83, opened 2018-10-03
Add support for sending and receiving messages via the ARGO Messaging Service

gregcorbett:
This PR allows the SSM to send messages via the ARGO Messaging Service. It:

- updates the config files to new, optional, settings that, when set, enable the AMS functionality
- allows sending to be authorized with a certificate and key, as opposed to with an AMS token
- provides documentation for setting up a new AMS sender and receiver
- provides documentation for migrating from the EGI message brokers to the AMS
- changes the log format when using AMS, to log the Argo Message ID - **this will require updating the logstash grok pattern used in our ELK monitoring before we deploy**


conf/receiver.cfg
```diff
@@ -19,6 +26,8 @@ capath: /etc/grid-security/certificates
 check_crls: false
 
 [messaging]
+# Project to which SSM will pull messages from. Uncomment and populate for AMS sending
```
tofu-rocketry: Move to sep AMS section

conf/receiver.cfg
```diff
@@ -1,3 +1,10 @@
+# SSM protocol/destination type options
+[SSM Type]
+# Either 'MQ-BROKER' for EGI Message Brokers or 'AMS' for Argo Messaging Service
```
tofu-rocketry: Collapse options to single one for user

ssm/ssm2.py
```diff
@@ -36,6 +36,8 @@
 import time
 import logging
 
+from argo_ams_library import ArgoMessagingService, AmsMessage, AmsException
```
tofu-rocketry: Prob only want either AMS or STOMP needed to be installed. Use SSL type check.

ssm/ssm2.py
```diff
@@ -55,9 +57,10 @@ class Ssm2(stomp.ConnectionListener):
     REJECT_SCHEMA = {'body': 'string', 'signer':'string?', 'empaid':'string?', 'error':'string'}
     CONNECTION_TIMEOUT = 10
 
-    def __init__(self, hosts_and_ports, qpath, cert, key, dest=None, listen=None,
-                 capath=None, check_crls=False, use_ssl=False, username=None, password=None,
-                 enc_cert=None, verify_enc_cert=True, pidfile=None):
+    def __init__(self, hosts_and_ports, qpath, cert, key, dest=None, listen=None, 
+                 capath=None, check_crls=False, use_ssl=False, username=None, password=None, 
+                 enc_cert=None, verify_enc_cert=True, pidfile=None, protocol="STOMP",
```
tofu-rocketry: Define as const

ssm/ssm2.py
```diff
@@ -311,6 +333,90 @@ def _send_msg(self, message, msgid):
             # If it fails, use the v3 metod signiture
             self._conn.send(to_send, headers=headers)
 
+    def pull_msg_rest(self):
```
tofu-rocketry: Pull out AMS and STOMP functions to own files.

bin/sender.py
```diff
@@ -62,40 +62,102 @@ def main():
         sys.exit(1)
     
     log = logging.getLogger('ssmsend')
-    
+
+    # Set defaults for MQ-BROKER only variables
+    use_ssl = None
+    # Set defaults for AMS only variables
+    token = ""
+    project = None
```
tofu-rocketry: Already ha default in SSM class

bin/sender.py
```diff
@@ -62,40 +62,102 @@ def main():
         sys.exit(1)
     
     log = logging.getLogger('ssmsend')
-    
+
+    # Set defaults for MQ-BROKER only variables
+    use_ssl = None
+    # Set defaults for AMS only variables
+    token = ""
```
tofu-rocketry: Look at defaulting this in init (currently goes into password)

tofu-rocketry: Note that empty string is needed as AMS only uses that to define abscence of token.

bin/receiver.py
```diff
@@ -24,6 +24,7 @@
 from ssm import __version__, set_up_logging, LOG_BREAK
 
 from stomp.exception import NotConnectedException
+from argo_ams_library import AmsConnectionException as NotConnectedException
```
tofu-rocketry: C.f. just accepting this error rather than aliasing.

bin/receiver.py
```diff
@@ -179,25 +239,31 @@ def main():
...
         # The message listening loop.
         while True:
+            try:
+                time.sleep(1)
+                if protocol == 'HTTPS':
+                    # We need to pull down messages as part of
+                    # this loop when using HTTPS.
+                    ssm.pull_msg_rest()
+
+                if i % REFRESH_DNS == 0:
```
tofu-rocketry: Poss doesn't need to be in try...except. Check if affects ELK.

ssm/ssm2.py
```diff
@@ -311,6 +333,90 @@ def _send_msg(self, message, msgid):
...
+                # sent via the SSM, we need to pull down that message
+                # to prevent it blocking the message queue.
+                log.debug("Message %s has no empaid.", msgid)
+                empaid = "N/A"
+            # get the message body
+            body = msg.get_data()
+
+            log.info('Received message. ID = %s, Argo ID = %s', empaid, msgid)
+
+            extracted_msg, signer, err_msg = self._handle_msg(body)
```
tofu-rocketry: This and bit below duplicated (to 405) from STOMP version. Should factor out to method and just use that. Take care of ack parts for AMS.

ssm/ssm2.py
```diff
@@ -342,14 +450,50 @@ def send_all(self):
...
+                while self._last_msg is None:
+                    if not self.connected:
+                        raise Ssm2Exception('Lost connection.')
+
+                log_string = "Sent %s" % msgid
+
+            elif self._protocol == "HTTPS" and self._dest_type == "AMS":
+                # Then we are sending to an Argo Messaging Service instance.
+                if text is not None:
+                    # First we sign the message
```
tofu-rocketry: Factor out to an AMS send message function.

conf/receiver.cfg
tofu-rocketry: Actually, I'm just going to change the name of the option and tweak the message.

conf/sender.cfg
```diff
@@ -1,3 +1,11 @@
+################################################################################
```
tofu-rocketry: Apply same changes from receiver.

ssm/ssm2.py
tofu-rocketry: For simplicity of the code, I think we should postpone making it an either/or choice and just require both for the moment.

bin/receiver.py
tofu-rocketry: Will leave it as there is the `send_ping` bit which could potentially raise an exception.

bin/sender.py
tofu-rocketry: Which is irrelevant as it still needs to be set to pass to the `Ssm2` constructor. However, most of these settings are fetched from config with exception handling if they're not present.

tofu-rocketry:
Pull request opened against GC's branch: gregcorbett/ssm#2

ssm/ssm2.py
tofu-rocketry: Deferred to #95.

ssm/ssm2.py
tofu-rocketry: Deferred to #95.

ssm/ssm2.py
tofu-rocketry: Deferred to #95.

tofu-rocketry:
I've rebased. There were conflicts with the changes from #72 and #93 which I've resolved. Hopefully I haven't made any mistakes. All reviews appreciated.

tofu-rocketry:
@gregcorbett, I can't add you as a reviewer as you opened the PR, but if you could also check my rebasing that would be good.

migrating_to_ams.md
```diff
@@ -0,0 +1,50 @@
...
+
+The sender configuration is usually found under `/etc/apel/sender.cfg`. Follow the steps below to migrate.
+
+1. Comment out `bdii` and `network`.
+1. Uncomment `host` and set it to `msg-devel.argo.grnet.gr`.
+1. Add the following as a new section at the top of your configuration:
+```
+[sender]
+# Either 'STOMP' for STOMP message brokers or 'AMS' for Argo Messaging Service
+protocol: STOMP
```
gregcorbett: this should be `protocol: AMS`, as the instruction says:
> Add the following as a new section at the top of your configuration:

tofu-rocketry: Good spot.

bin/receiver.py
```diff
@@ -103,39 +104,84 @@ def main():
...
-        log.error('Could not connect to LDAP server: %s', e)
-        log.error('System will exit.')
-        log.info(LOG_BREAK)
-        sys.exit(1)    
-    
+
+        # Attempt to configure AMS specific variables.
+        try:
+            token = cp.get('messaging', 'token')
+            project = cp.get('messaging', 'project')
```
gregcorbett: This should be `'ams_project'`

bin/sender.py
```diff
@@ -60,42 +60,96 @@ def main():
...
-        log.error('Could not connect to LDAP server: %s', e)
-        log.error('System will exit.')
-        log.info(LOG_BREAK)
-        print 'SSM failed to start.  See log file for details.'
-        sys.exit(1)
-        
+
+        # Attempt to configure AMS project variable.
+        try:
+            project = cp.get('messaging', 'project')
```
gregcorbett: This should be `'ams_project'`

migrating_to_ams.md
tofu-rocketry: Sorted.

bin/receiver.py
tofu-rocketry: Corrected.

bin/sender.py
tofu-rocketry: Corrected.

tofu-rocketry:
Any final comments before I merge this?

I'll add the argo-ams-library to the SPEC an setup.py when I'm on the release branch.

gregcorbett:
I am trying to do a test send/receive, currently hitting an error around the `use_ssl` variable.

bin/sender.py
```diff
@@ -130,16 +184,19 @@ def main():
...
-                   capath=cp.get('certificates', 'capath'),
-                   enc_cert=server_cert,
-                   verify_enc_cert=verify_server_cert)
-        
+                      cp.get('messaging', 'path'),
+                      path_type=path_type,
+                      cert=cp.get('certificates', 'certificate'),
+                      key=cp.get('certificates', 'key'),
+                      dest=cp.get('messaging', 'destination'),
+                      use_ssl=use_ssl,
```
gregcorbett: if using `Ssm2.AMS_MESSAGING`, `use_ssl` doesn't get set so this using it here causes an error.

tofu-rocketry: My bad. I've re-added setting it to `None`, but only when using AMS this time, for both sender and receiver.

gregcorbett:
I can get both the `sender` and `receiver` to the point where the broker gives me a `401`.

PR #93, opened 2019-03-04
Mock out external LDAP search calls in tests

### LINKED ISSUE #85: Mock out call to LDAP server in tests
Unit tests are fragile as the test below makes a call to the CERN BDII. This should be mocked out.

https://github.com/apel/ssm/blob/6376e3105dfbc4282d24a4342774269d65f44558/test/test_brokers.py#L67

tofu-rocketry:
Resolves #85 while tidying up the code a bit.

### DIFF
--- a/test/test_brokers.py
+++ b/test/test_brokers.py
@@ -1,38 +1,26 @@
-'''
-   Copyright (C) 2012 STFC
-
-   Licensed under the Apache License, Version 2.0 (the "License");
-   you may not use this file except in compliance with the License.
-   You may obtain a copy of the License at
-
-       http://www.apache.org/licenses/LICENSE-2.0
-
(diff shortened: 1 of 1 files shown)


test/test_brokers.py
```diff
@@ -73,6 +78,45 @@ def test_fetch_brokers(self):
         if '.' not in host:
             self.fail("Didn't get a hostname from fetch_brokers()")
 
+    def _mocked_search(*args, **kwargs):
+        """Return values to mocked search call based on input."""
+
+        if (
```
gregcorbett: atleast it's PEP8 compliant....

tofu-rocketry: Yeah, it's not the nicest, but the LDAP queries aren't the shortest!

PR #97, opened 2019-05-17
OpenSSL checkend fix

### LINKED ISSUE #96: openssl 1.1 problem
With openssl 1.1, the crypto.py complains on syntax error in this line:
https://github.com/apel/ssm/blob/055aeb2e8e90e81605ec7c6b6e78d7f66069118f/ssm/crypto.py#L221
```
# openssl x509 -checkend -noout -in /etc/grid-security/hostcert.pem
x509: Can't parse "-noout" as a number
x509: Invalid number "-noout" for -checkend
x509: Use -help for summary.
```

adding argument to checkend, eg "1" works as expected:
```
# openssl x509 -checkend 1 -noout -in /etc/grid-security/hostcert.pem
Certificate will not expire
```

If the line is replaced by 
`args = ['openssl', 'x509', '-checkend', '1', '-noout', '-in', certpath]`
it works for both openssl 1.0 and openssl 1.1

Could be this fixed?

tofu-rocketry:
Resolves #96. Note that this will be merged into the release branch.

`-checkend` should be followed by a number that defines how many seconds ahead `openssl x509` should check for cert expiry, otherwise later versions of OpenSSL raise syntax errors (treating next arg as the number). Set it to a day as certs should really be renewed before that considering the amount of time it can take a message to traverse the messaging system.

Other minor changes to exception messages and tests in support of this.

### DIFF
--- a/ssm/crypto.py
+++ b/ssm/crypto.py
@@ -214,11 +214,16 @@ def decrypt(encrypted_text, certpath, keypath):
 
 
 def verify_cert_date(certpath):
-    """Return True if certifcate is 'in date', otherwise return False."""
+    """Check that certificate hasn't expired and won't expire within 24 hours.
+
+    Return True if certifcate is 'in date', otherwise return False.
+    """
     if certpath is None:
         raise CryptoException('Invalid None argument to verify_cert_date().')
 
-    args = ['openssl', 'x509', '-checkend', '-noout', '-in', certpath]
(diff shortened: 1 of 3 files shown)


tofu-rocketry:
> LGTM, feels like it might be worth squashing the commits upon merge.

Thanks. I think I'd prefer to have a merge commit for clarity as we don't often do fixes like this on release branches.

PR #99, opened 2019-07-31
Raise CryptoException if signed message verification fails (for whatever reason).

### LINKED ISSUE #39: crypto.verify returns a certificate/subject pair even if a signed message has been altered
Discovered during development of #38.

`crypto.verify` returns a certificate/subject pair if the signed message has been tampered with after signing and hides the fact underlying openssl verification fails.

This is because :

- `crypto.get_signer_cert()` returns the signers certificate even if the signature is invalid e.g. the message body is tampered with after signing (line [148](https://github.com/apel/ssm/blob/dev/ssm/crypto.py#L148))
- The openssl command and `p1.communicate(signed_text)` can still set `message` to the message content, even if `error` is also set (lines [156](https://github.com/apel/ssm/blob/dev/ssm/crypto.py#L156) and [159](https://github.com/apel/ssm/blob/dev/ssm/crypto.py#L159))
- `crypto.verify` then returns the message body and signer certificate subject regardless of `error` (line [182](https://github.com/apel/ssm/blob/dev/ssm/crypto.py#L182))

The openssl error is logged, but the tampered message is then treated as verified.

gregcorbett:
One such reason is the message has been tampered with after being signing. Therefore this resolves #39.

### DIFF
--- a/ssm/crypto.py
+++ b/ssm/crypto.py
@@ -179,9 +179,8 @@ def verify(signed_text, capath, check_crl):
     if "Verification successful" in error:
         log.debug(error)
     else:
-        log.warn(error)
         raise CryptoException(
-            "Failed to verify the signed message."
+            "Possible tampering. See OpenSSL error: %s" % error
         )
 
     subj = get_certificate_subject(signer)
--- a/ssm/crypto.py
+++ b/ssm/crypto.py
@@ -181,7 +181,7 @@ def verify(signed_text, capath, check_crl):
     else:
         log.warn(error)
         raise CryptoException(
-            "Failed to verify the signed message, see the log for details."
+            "Failed to verify the signed message."
         )
 
     subj = get_certificate_subject(signer)
--- a/ssm/crypto.py
+++ b/ssm/crypto.py
@@ -180,6 +180,9 @@ def verify(signed_text, capath, check_crl):
         log.debug(error)
     else:
         log.warn(error)
+        raise CryptoException(
+            "Failed to verify the signed message, see the log for details."
+        )
 
     subj = get_certificate_subject(signer)
     return body, subj
@@ -329,4 +332,3 @@ def get_signer_cert(signed_text):
         log.error(error)
 
     return certstring
-
--- a/test/test_crypto.py
+++ b/test/test_crypto.py
@@ -281,7 +281,12 @@ class TestEncryptUtils(unittest.TestCase):
         tampered_message = signed_message.replace(MSG, "Spam")
 
         # Verifying the orignal, un-tampered message should be fine.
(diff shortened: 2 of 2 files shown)


gregcorbett:
This requires the release-2.4.0 branch to be merged first as that contains the fix for the CI build.

test/test_crypto.py
```diff
@@ -275,6 +275,18 @@ def test_verify_cert(self):
         except CryptoException:
             pass
 
+    def test_message_tampering(self):
+        """Test that a tampered message is not accepted as valid."""
+        signed_message = sign(MSG, TEST_CERT_FILE, TEST_KEY_FILE)
+        tampered_message = signed_message.replace(MSG, "Spam")
+
+        # Verifying the orignal, un-tampered message should be fine.
+        verify(signed_message, TEST_CA_DIR, False)
```
tofu-rocketry: Should either do a check on the output of `verify` or remove this line - it's not really doing anything in a test case.

ssm/crypto.py
```diff
@@ -180,6 +180,9 @@ def verify(signed_text, capath, check_crl):
         log.debug(error)
     else:
         log.warn(error)
+        raise CryptoException(
+            "Failed to verify the signed message, see the log for details."
```
tofu-rocketry: Do we need to say to check the log? Where would this message appear? We're already dumping out the error message in the previous line.

tofu-rocketry:
Rebased.

test/test_crypto.py
gregcorbett: Yes we should. I have added a check on both the message returned and the DN of the signer.

ssm/crypto.py
gregcorbett: I think the "see the log" message would only appear in the logs, so we don't need to say it. Removed.

gregcorbett:
I further rebased to "get over" the whitespace changes from #102

tofu-rocketry:
It's not a problem, but I had already rebased it past @jrha's changes.

ssm/crypto.py
```diff
@@ -180,6 +180,9 @@ def verify(signed_text, capath, check_crl):
         log.debug(error)
     else:
         log.warn(error)
+        raise CryptoException(
+            "Failed to verify the signed message."
```
tofu-rocketry:
```suggestion
            "Possible tampering. See OpenSSL error: %s" % error
```
This message should probably contain the full error as the error handling when this is called already says pretty much this:

https://github.com/apel/ssm/blob/a75ded59967e24f694429760bdce960467dedaa3/ssm/ssm2.py#L329

PR #114, opened 2020-01-22
Build update

### LINKED ISSUE #113: Remove restriction on python-daemon
Introduced in #75 and raised in this [GGUS ticket](https://ggus.eu/index.php?mode=ticket_info&ticket_id=144885), this restriction should now hopefully not be needed.

See also #80.

tofu-rocketry:
Resolves #113.

This PR removes the Python 2.6 build and the dependency version constraints that were required to maintain compatibility. It does however constrain the version of python-daemon to maintain Python 2.7 compatibility.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,14 +1,12 @@
+os: linux
 language: python
 python:
   - "2.7"
   - "3.7"
-matrix:
+jobs:
   allow_failures:
     - python: "3.7"
   fast_finish: true
-
-# Pin Ubuntu to Trusty for the moment for Python 2.6 support
-dist: trusty
 
 # Cache the dependencies installed by pip
 cache: pip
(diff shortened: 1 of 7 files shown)


gregcorbett:
README.md specifies `stomp.py` versions "between 3.1.1 (inclusive) and 5.0.0 (exclusive)", but `apel-ssm.spec`, `requirements.txt ` and `setup.py` only constrains `stomp.py` to (strictly) below 5.0.0.

Can a lower bound be specified in those files as well as an upper?

tofu-rocketry:
> Can a lower bound be specified in those files as well as an upper?

Only in the latest versions of RPM (I didn't check for setuptools), so I thought I'd go for the constraint that's most relevant at the moment.

The README then covers that fact that it's not compatible.

Issue #91, opened 2018-12-11
Things to tidy up


tofu-rocketry:
While working on AMS integration, I've noticed the following:

- Remove password and username as they aren't used
- use_ssl defaults to False in Ssm2 init. Switch to True and check it defaults True everywhere else
- check_crls: false in receiver config - investigate if this should be true
- loads of PEP8 violations

tofu-rocketry:
For licences and docstrings, `apgents.py` has the layout I prefer.

PR #115, opened 2020-01-24
Drop Python 2.6, add Python 3 test compatibility

tofu-rocketry:
Partially addresses #91.

**N.B. Removes support for Python 2.6.** So we'll need to go up a major (!) version number for the next release.

This pull request adds basic Python 3 compatibility such that the Travis tests can run successfully. There are certainly other things that will need fixing to ensure SSM actually works with Python 3 (including `optparse` deprecation).

References for `quopri` changes:
[Python 2](https://docs.python.org/2/library/quopri.html)
[Python 3](https://docs.python.org/3/library/quopri.html)

### DIFF
--- a/ssm/crypto.py
+++ b/ssm/crypto.py
@@ -175,6 +175,11 @@ def verify(signed_text, capath, check_crl):
         body = base64.decodestring(body)
     # otherwise, plain text
 
+    # In Python 3, decodestring() returns bytes so decode to a string while
+    # Python 2 compatability is still required.
+    if not isinstance(body, str):
+        body = body.decode()
+
     # 'openssl smime' returns "Verification successful" to standard error. We
     # don't want to log this as an error each time, but we do want to see if
     # there's a genuine error.
--- a/ssm/crypto.py
+++ b/ssm/crypto.py
@@ -319,7 +319,7 @@ def get_certificate_subject(certstring):
 
 
 def get_signer_cert(signed_text):
-    """Read the signer's certificate from the signed specified message."""
+    """Return the signer's certificate from the signed specified message."""
     # This ensures that openssl knows that the string is finished.
     # It makes no difference if the signed message is correct, but
     # prevents it from hanging in the case of an empty string.
--- a/ssm/crypto.py
+++ b/ssm/crypto.py
@@ -314,22 +314,22 @@ def get_certificate_subject(certstring):
 
 
 def get_signer_cert(signed_text):
-    '''
-    Read the signer's certificate from the signed specified message, and return the
-    certificate string.
-    '''
+    """Read the signer's certificate from the signed specified message."""
     # This ensures that openssl knows that the string is finished.
     # It makes no difference if the signed message is correct, but
     # prevents it from hanging in the case of an empty string.
     signed_text += '\n\n'
 
-    p1 = Popen(['openssl', 'smime', '-pk7out'],
-               stdin=PIPE, stdout=PIPE, stderr=PIPE)
-    p2 = Popen(['openssl', 'pkcs7', '-print_certs'],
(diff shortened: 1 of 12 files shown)


ssm/crypto.py
```diff
@@ -66,15 +67,15 @@ def check_cert_key(certpath, keypath):
         return False
 
     p1 = Popen(['openssl', 'x509', '-noout', '-modulus'],
-               stdin=PIPE, stdout=PIPE, stderr=PIPE)
+               stdin=PIPE, stdout=PIPE, stderr=PIPE, universal_newlines=True)
```
gregcorbett: Is `universal_newlines=True` added for backwards compatibility with Python 2.6? Either way, would be useful to add a comment here as to why it's needed.

ssm/crypto.py
```diff
@@ -310,22 +319,22 @@ def get_certificate_subject(certstring):
 
 
 def get_signer_cert(signed_text):
-    '''
-    Read the signer's certificate from the signed specified message, and return the
-    certificate string.
-    '''
+    """Read the signer's certificate from the signed specified message."""
```
gregcorbett: Why lose the reference to what the method returns?

```suggestion
    """Returns the signer's certificate from the signed specified message."""
```

ssm/message_directory.py
```diff
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 """This module contains the MessageDirectory class."""
+from __future__ import print_function
```
gregcorbett: Is this import needed? It doesn't seem to be used in this file/class?

tofu-rocketry: I wanted to avoid regression if a print statement gets added.

ssm/crypto.py
tofu-rocketry: I'd say it's actually added for forward compatibility with Python 3. Python 2.7 doesn't require it. I used `universal_newlines` rather than the later `text` argument, which only exists in later versions of Pyhton 3, for backwards compatibility. Comments could be useful, though I was wary of littering the code with comments for a fairly simple change.

ssm/crypto.py
tofu-rocketry: I was just trying to remove redundant information, but yes, I should have kept the info on what it returns. It should be in the imperative mood though.

ssm/message_directory.py
gregcorbett: In that case wouldn't the Python 3.7 Travis build fail?

And if a Python 3 style print function gets added to the code later, wouldn't the Python 2.7 build fail if this line was omitted where it was then needed.

ssm/crypto.py
gregcorbett: I've updated the suggestion above (can't seem to add a suggestion in a comment under a comment)

ssm/message_directory.py
tofu-rocketry: > In that case wouldn't the Python 3.7 Travis build fail?

Only if there's a test that executes the code.

> And if a Python 3 style print function gets added to the code later, wouldn't the Python 2.7 build fail if this line was omitted where it was then needed.

It wouldn't as Python 2.7 is able to ignore the parentheses and it would treat it as just a print statement, but I'd rather it was treated it the same as Python 3, hence the import.

The general idea is to make whatever code is in the file behave the same under Python 2.7 or 3.

gregcorbett: > Python 2.7 is able to ignore the parentheses and it would treat it as just a print statement

Ah yes, it will. Fair enough.

tofu-rocketry:
I accept your ambivalence. :balance_scale:

PR #119, opened 2020-02-06
95 refactor

### LINKED ISSUE #95: AMS support: optional modules and extra refactoring
It should be optional which module is installed - `argo-ams-library` or `stomp.py`. See also #80.

Additionally, these are the outstanding bits of refactoring from #83 that I decided to postpone as they're only nice-to-haves.

- Pull out AMS and STOMP functions to own files. https://github.com/apel/ssm/pull/83/files#diff-591de25f6e8834558a6bf8d6462ad266R336

- This and bit below duplicated (to 405) from STOMP version. Should factor out to method and just use that. Take care of ack parts for AMS. https://github.com/apel/ssm/pull/83/files#diff-591de25f6e8834558a6bf8d6462ad266R378

- Factor out to an AMS send message function. https://github.com/apel/ssm/pull/83/files#diff-591de25f6e8834558a6bf8d6462ad266R469

tofu-rocketry:
This resolves #95 by addressing most of the items in the issue.

- `argo-ams-library` has been made optional.
- Duplicated code between AMS and STOMP methods has been factored out to `_save_msg_to_queue`.
- AMS send message function has been factored out to `_send_msg_ams`.

It doesn't pull out AMS and STOMP functions to their own files, but that's a larger thing that is probably dependant on `stomp.py` being made optional (but it has a lot of low level integration in the code). Follow up issue in #120.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -51,10 +51,11 @@ def main():
           download_url='https://github.com/apel/ssm/releases',
           license='Apache License, Version 2.0',
           install_requires=[
-              'stomp.py<5.0.0', 'python-ldap', 'argo-ams-library',
+              'stomp.py<5.0.0', 'python-ldap',
           ],
           extras_require={
-              'python-daemon': ['python-daemon'],
+              'AMS': ['argo-ams-library'],
+              'daemon': ['python-daemon'],
               'dirq': ['dirq'],
           },
           packages=find_packages(exclude=['bin', 'test']),
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ The Python STOMP library (N.B. versions between 3.1.1 (inclusive) and 5.0.0
 (exclusive) are currently supported)
 * `yum install stomppy`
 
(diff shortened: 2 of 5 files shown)


setup.py
```diff
@@ -51,10 +51,11 @@ def main():
           download_url='https://github.com/apel/ssm/releases',
           license='Apache License, Version 2.0',
           install_requires=[
-              'stomp.py<5.0.0', 'python-ldap', 'argo-ams-library',
+              'stomp.py<5.0.0', 'python-ldap',
           ],
           extras_require={
-              'python-daemon': ['python-daemon'],
+              'AMS': ['argo-ams-library'],
+              'daemon': ['python-daemon'],
```
gregcorbett: Changing things?! I guess as this is milestone'd for 3.0.0 we can go crazy :p!

tofu-rocketry: The dictionary key is meant to be the name of the optional "feature" of your software rather than just the name of the package.

PR #121, opened 2020-02-06
Sleep tweaks

tofu-rocketry:
- Fix regression from 65971ae
   - The sleep was intended to limit the speed of the while loop that waited
for the STOMP message to be accepted, but this got moved outside the
while loop when AMS support was added, which does not seem sensible.
-  Speed up receiver loop
   - Speed up receiver loop by decreasing sleep as pulling a message only
once a second doesn't seem very efficient. The value of REFRESH_DNS
needs is increassed ten-fold to balance out the change in sleep.

### DIFF
--- a/bin/receiver.py
+++ b/bin/receiver.py
@@ -249,13 +249,13 @@ def main():
         # The message listening loop.
         while True:
             try:
-                time.sleep(1)
+                time.sleep(0.1)
                 if protocol == Ssm2.AMS_MESSAGING:
                     # We need to pull down messages as part of
                     # this loop when using AMS.
                     ssm.pull_msg_ams()
 
-                if i % REFRESH_DNS == 0:
+                if i % (REFRESH_DNS * 10) == 0:
                     log.info('Refreshing valid DNs and then sending ping.')
                     dns = get_dns(options.dn_file)
                     ssm.set_dns(dns)
--- a/ssm/ssm2.py
+++ b/ssm/ssm2.py
@@ -485,6 +485,8 @@ class Ssm2(stomp.ConnectionListener):
                 while self._last_msg is None:
                     if not self.connected:
                         raise Ssm2Exception('Lost connection.')
+                    # Small sleep to avoid hammering the CPU
+                    time.sleep(0.01)
 
                 log_string = "Sent %s" % msgid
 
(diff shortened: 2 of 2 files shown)


ssm/ssm2.py
```diff
@@ -485,6 +485,7 @@ def send_all(self):
                 while self._last_msg is None:
                     if not self.connected:
                         raise Ssm2Exception('Lost connection.')
+                    time.sleep(0.1)
```
gregcorbett: Maybe a comment as to why the `sleep` is needed?

bin/receiver.py
```diff
@@ -249,13 +249,13 @@ def main():
...
             try:
-                time.sleep(1)
+                time.sleep(0.1)
                 if protocol == Ssm2.AMS_MESSAGING:
                     # We need to pull down messages as part of
                     # this loop when using AMS.
                     ssm.pull_msg_ams()
 
-                if i % REFRESH_DNS == 0:
+                if i % (REFRESH_DNS * 10) == 0:
```
gregcorbett: Why not just increase the value of REFRESH_DNS by ten?

tofu-rocketry: As it's defined as being in seconds and this follows the same pattern used somewhere else. I guess changing the definition might be better, but I was too lazy to look at where `REFRESH_DNS` was defined and to also check that it wasn't being used anywhere else. :stuck_out_tongue_closed_eyes:

ssm/ssm2.py
tofu-rocketry: I believe the point is that when you are running a `while` loop you don't want to hammer out the retries as that's a waste of CPU resources. I'll come up with a comment.

tofu-rocketry: Actually doing some testing, only a very small sleep is required to avoid hogging CPU, so I might reduce it.

tofu-rocketry: ...but we may then need to keep the sleep below in for rate limiting purposes. Testing it against the brokers is probably the easiest way to see the difference.

bin/receiver.py
tofu-rocketry: Will leave as-is.

ssm/ssm2.py
tofu-rocketry: Comment added and sleep reduced.

gregcorbett (approved):

PR #125, opened 2020-04-04
generalize certificate and a key matching verification to admit both RSA and EC keys

dealfonso:
Resolves #124.

Changes in the verification of key/cert matching, to correct the bug in issue #124 

The new verification algorithm makes use of suggestions in https://security.stackexchange.com/a/143981 to compare the public key extracted from both the certificate and the private key.

```
openssl x509 -pubkey -in mydomain-ecc.crt -noout
openssl pkey -pubout -in mydomain-ecc.key
```

PR #132, opened 2020-10-08
Tidy up docstrings and some other style bits

tofu-rocketry:
This will have some mild conflicts with #125 in ssm/crypto.py due to the line re-flowing.


jrha (approved):
:+1:

PR #135, opened 2020-12-01
Add retry=3 to AMS methods for reliability

### LINKED ISSUE #134: Add retries to ams methods
AMS library >= 0.5 includes support for retries, but defaults to not retrying (!)

We need to add these in to our calls to `publish` and `pull_sub` in `ssm2.py` . Examples here: https://github.com/ARGOeu/argo-ams-library/blob/v0.5.2-1/examples/retry.py

Should possibly also retry with `ack_sub`.

tofu-rocketry:
Resolves #134.

This is backwards compatible as the AMS methods all have a **kwargs argument in the previous versions.


gregcorbett (approved):
LGTM :)

PR #143, opened 2020-12-14
Add another urllib logging tweak

tofu-rocketry:
Something seems to have changed behind the scenes with has required tweaking the logging for urllib3 again as we did in #105. This could have been a change in AMS, its library, or something peculiar to our installation on our prod machine (this behaviour wasn't seen on pre-prod).

### DIFF
--- a/ssm/ssm2.py
+++ b/ssm/ssm2.py
@@ -168,14 +168,18 @@ class Ssm2(stomp.ConnectionListener):
                     raise Ssm2Exception('Failed to verify server certificate %s against CA path %s.'
                                         % (self._enc_cert, self._capath))
 
-        # If the overall SSM log level is info, we want to only
+        self._set_external_logging_levels()
+
+    def _set_external_logging_levels(self):
+        """Tweak the logging of dependencies to better match SSM verbosity."""
+        # If the overall SSM log level is INFO, we want to only
         # see entries from stomp.py and connectionpool at WARNING and above.
         if getLogger("ssm.ssm2").getEffectiveLevel() == INFO:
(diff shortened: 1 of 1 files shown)


ineilson (approved):
Looks ok to me. Your comment in another thread about using a loop would not, to me, make it easier to read at this stage.

ineilson (commented):
Dumb question: What happens if getEffectiveLevel is not INFO or DEBUG? Is the default state something else and guaranteed? Else, could just force INFO level default and use single if around DEBUG to be certain its set as you want.

tofu-rocketry:
> Dumb question: What happens if `getEffectiveLevel` is not INFO or DEBUG? Is the default state something else and guaranteed? Else, could just force INFO level default and use single if around DEBUG to be certain its set as you want.

There are the levels of `WARNING`, `ERROR` and `CRITICAL`, but I'd say those options would be a bit unusual to set in the config. It's not something I've used so the behaviour is somewhat undefined.

> ...use single if around DEBUG...

Not sure if that's a typo, but generally for us I think it would make more sense the other way round - to force WARNING on the external modules in most cases and only set INFO for them if SSM is set to DEBUG. I seem to recall that DEBUG was super noisy for `urllib3`.

tofu-rocketry:
Oh, just figured out why we're not worried about the other logging levels - loggers inherit the levels, which means that so long as you're at WARNING or above, the external modules should be at WARNING or above and so not annoying you.

ineilson (approved):

jrha (approved):
Much improved :+1:

PR #149, opened 2021-03-10
Messaging agents refactor

tofu-rocketry:
As a first step to fixing apel/apel#230, I've done an initial refactor of the SSM sender and receiver executables so that most code is now in `agents.py`.

There's probably more changes to the tests I need to do.


gregcorbett (dismissed):
Looks good to me, definitely an improvement :)

tofu-rocketry:
Testing raised some issues, mostly with the receiver, but it's meant that I've improved things for both.

gregcorbett (approved):
LGTM

PR #162, opened 2021-04-28
Release 3.2.1 to dev

tofu-rocketry:
There's one change on top of the normal ones - an extra note in the README file as the docs should match how the dependencies are now set up as a result of this update.


jrha (approved):
LGTM

PR #165, opened 2021-06-18
Add RPM build workflow for tags

tofu-rocketry:
Tested in my fork. Pushing a tag correctly invokes the build.

I'm leaving the RPMs as artefacts on the build for now (rather than pushing them to the release), as I'll want to compare them with the existing process next time I do a release.


tofu-rocketry:
The scheduled cron build does seem to have run in my fork.

.github/workflows/build-pkgs.yml
```diff
@@ -0,0 +1,54 @@
...
+    - name: Install rpmlint
+      run: |
+        sudo apt-get update
+        sudo apt-get install rpmlint
+    - uses: actions/checkout@v2
+      with:
+        # Get all branches and tags so the latest tag can be found for VERSION
+        fetch-depth: 0
+    - name: Set VERSION
+      run: echo "VERSION=`echo $(git describe --tags $(git rev-list --tags --max-count=1))`" >> $GITHUB_ENV
```
gregcorbett, line 24: Can you get the tag that is being built from the [GitHub context](https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#github-context)? Specifically `github.ref`.

gregcorbett: No, as that would give `refs/tags/<tag_name>` rather than just `<tag_name>`.

PR #166, opened 2021-06-23
GitHub actions for SSM

RoseECooper:
Add in ssm_actions file to replace the travis file, currently skips some problematic tests in the test_crypto.py file.


tofu-rocketry:
I've reproduced the test_crypto errors on an Ubuntu VM. Should check if it's Ubuntu specific (i.e. see if they've broken on RHEL too).

tofu-rocketry:
I've reproduced it on SL7 by forcing the code to use OpenSSL 1.1.x whereas it usually uses 1.0.x. That seems to be the difference.

tofu-rocketry:
The GH Actions tests themselves are failing due to the OpenSSL issue #167 , but I've resurrected the Travis tests for now on travis-ci.com and I'd kinda like to have this issue highlighted with the failing Action for now.

PR #169, opened 2021-06-28
Specify more specific rpms for build

tofu-rocketry:
Not every build system will have access to fedora-packager, but only a
subset of what's included with that is actually required.


gregcorbett (approved):
LGTM

jrha (approved):
:+1:

PR #173, opened 2021-07-23
Remove line 311 to allow relative dns file path

RoseECooper:


ssm/agents.py
```diff
@@ -308,7 +308,6 @@ def run_receiver(protocol, brokers, project, token, cp, log, dn_file):
 
                 if i % (REFRESH_DNS * 10) == 0:
                     log.info('Refreshing valid DNs and then sending ping.')
-                    dns = get_dns(dn_file, log)
```
tofu-rocketry, line 311: What's the thinking behind this? Removing this line is just going to prevent the DNs being updated. `dn_file` needs to be re-read which is done by calling `get_dns()`. It's either that function that needs fixing, or the setting of the `dn_file` value.

RoseECooper: Ahh okay, I misunderstood what was happening here so I'll add that line back in and work on either fixing the function or the `dn_file`.

PR #175, opened 2021-08-13
Remove python-daemon as a requirement 

### LINKED ISSUE #80: python-daemon should not be listed as a requirement as it is not a hard requirement.
Currently, the `python-daemon` package is an optional requirement in [setup.py](https://github.com/apel/ssm/blob/dev/setup.py#L61) but a hard requirement in [apel-ssm.spec](https://github.com/apel/ssm/blob/dev/apel-ssm.spec#L24). As there's no consistent way of handling optionals across packaging systems, `python-daemon` should just be left out.

We need exception handling for the case when the package is missing and an attempt is made to use the parts of our software that need it. We should also research the best way to do this.

See #70 for the original question.

gregcorbett:
Resolves #80

As there's no consistent way of handling optional packages across packaging systems, python-daemon should just be left out.

### DIFF
--- a/ssm/agents.py
+++ b/ssm/agents.py
@@ -18,7 +18,6 @@
 
 from __future__ import absolute_import, division, print_function, unicode_literals
 
-from daemon import DaemonContext
 import logging
 import ldap
 import os
@@ -29,6 +28,13 @@ try:
     import ConfigParser
 except ImportError:
     import configparser as ConfigParser
+
+try:
+    from daemon import DaemonContext
+except ImportError:
+    # A error is logged and the receiver exits later if DaemonContext is
+    # requested but not installed.
+    DaemonContext = None
 
 from stomp.exception import NotConnectedException
 try:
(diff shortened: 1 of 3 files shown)


tofu-rocketry:
That's the rebase done.

ssm/agents.py
```diff
@@ -259,6 +265,13 @@ def run_sender(protocol, brokers, project, token, cp, log):
 
 def run_receiver(protocol, brokers, project, token, cp, log, dn_file):
     """Run Ssm2 as a receiver daemon."""
+    if DaemonContext is None:
```
tofu-rocketry, line 268: I moved to this the equivalent place in `agents.py` following the rebase. Before the rebase, it would have been nice to run this earlier in `receiver.py` (to avoid some of the setup sets which would be pointless if python-daemon isn't available), but that doesn't make sense following the rebase so here it shall stay!

jounaidr (commented):
LGTM

tofu-rocketry (approved):

Issue #155, opened 2021-03-19
Handle AMS message too large error gracefully


tofu-rocketry:
Currently you get this:

```
2021-03-15 15:10:05,560 - ssmsend - ERROR - Unexpected exception in SSM: {'status': u'INVALID_ARGUMENT', 'status_code': 413, 'error': 'While trying the [topic_publish]: Message size is too large'}
2021-03-15 15:10:05,560 - ssmsend - ERROR - Exception type: <class 'argo_ams_library.amsexceptions.AmsServiceException'>
```

As reported [here](https://ggus.eu/index.php?mode=ticket_info&ticket_id=150992).

PR #179, opened 2021-09-13
156 1mb error handling

### LINKED ISSUE #156: Warn and handle if AMS messages are > 1 MB
Related to #74 as it could/should be implemented alongside that.

AMS has a limit of 1MB for messages as that is what it is optimised around. SSM should check the size of messages (and any associated wrapping) before sending them to check they won't hit this limit, and ideally handle a fail gracefully. #74 could then be used to split the message into manageable parts.

Should also consider what to do if any messages can't be sent - send what you can, or fail completely? Probably send what you can.

jounaidr:
Resolves #156


gregcorbett:
> Think I'll extract the basic changes for #155 for the next release and save the bits for #156 for later.

Do the commits left in this PR represent just the changes needed for #156 currently?

tofu-rocketry:
> Do the commits left in this PR represent just the changes needed for #156 currently?

No, I've not split it out yet.

PR #181, opened 2021-10-07
151 DNs file path on command line

### LINKED ISSUE #151: DNs file on command line can't be relative
Unlike the config option `-c` (and maybe others), the DNs file option `-d` for the receiver seems to require an absolute path. It should allow a relative path like the other options.

Also check the other options work the same

RoseECooper:
Resolves #151 
Allows a relative file path for the DNs file from the command line.

### DIFF
--- a/bin/receiver.py
+++ b/bin/receiver.py
@@ -47,7 +47,7 @@ def main():
 
     (options, unused_args) = op.parse_args()
 
-    # Give dn_file absolute path
+    # Absolute file path required when refreshing dn_file, relative path resulted in an error. 
     options.dn_file=os.path.abspath(options.dn_file)
 
     cp = ConfigParser.ConfigParser({'use_ssl': 'true'})
--- a/bin/receiver.py
+++ b/bin/receiver.py
@@ -47,6 +47,9 @@ def main():
 
     (options, unused_args) = op.parse_args()
 
+    # Give dn_file absolute path
+    options.dn_file=os.path.abspath(options.dn_file)
+
     cp = ConfigParser.ConfigParser({'use_ssl': 'true'})
     cp.read(options.config)
 
--- a/bin/receiver.py
+++ b/bin/receiver.py
@@ -48,7 +48,7 @@ def main():
     (options, unused_args) = op.parse_args()
 
     # Absolute file path required when refreshing dn_file, relative path resulted in an error. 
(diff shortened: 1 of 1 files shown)


ssm/agents.py
```diff
@@ -362,6 +362,6 @@ def get_dns(dn_file, log):
     # If no valid DNs, SSM cannot receive any messages.
     if len(dns) == 0:
         raise Ssm2Exception('No valid DNs found in %s.  SSM will not start' % dn_file)
-
+
```
tofu-rocketry, line 365: Can you remove/undo the whitespace changes in this file?

bin/receiver.py
```diff
@@ -47,6 +47,9 @@ def main():
 
     (options, unused_args) = op.parse_args()
 
+    # Give dn_file absolute path
```
tofu-rocketry, line 50: That's the 'what', but the code tells you that. The comment should say 'why'.

tofu-rocketry (changes requested):
Ah, it was just the refreshes that had the issue - explains why it was a bit tricky to reproduce.

Just some tidying needed.

ssm/agents.py
```diff
@@ -365,3 +365,4 @@ def get_dns(dn_file, log):
 
     log.debug('%s DNs found.', len(dns))
     return dns
+
```
tofu-rocketry, line 368: There were some spaces hiding on the last line.

```suggestion

```

bin/receiver.py
```diff
@@ -47,6 +47,9 @@ def main():
 
     (options, unused_args) = op.parse_args()
 
+    # Absolute file path required when refreshing dn_file, relative path resulted in an error. 
+    options.dn_file=os.path.abspath(options.dn_file)
```
tofu-rocketry, line 51: Spaces around operators is the usual Python style. 

```suggestion
    options.dn_file = os.path.abspath(options.dn_file)
```

ssm/agents.py
```diff
@@ -364,4 +364,4 @@ def get_dns(dn_file, log):
         raise Ssm2Exception('No valid DNs found in %s.  SSM will not start' % dn_file)
 
     log.debug('%s DNs found.', len(dns))
-    return dns
+    return dns
```
tofu-rocketry, line 367:
```suggestion
    return dns

```

tofu-rocketry (approved):

PR #201, opened 2022-07-06
Set GH Action unit test matrix to not fail fast

tofu-rocketry:
- This means that every job will run, whatever the result of any other.


gregcorbett (approved):
LGTM

PR #203, opened 2022-07-22
176 remove log config

### LINKED ISSUE #176: Deprecate log config
logging.cfg causes a number of problems, not least that it doesn't seem to work and gives unexpected results if it's included without realising it. We should deprecate it and any related functionality (at least part of logging_helper and how it's handled in the executables).

Letizia97:
Resolves #176


bin/receiver.py
```diff
@@ -38,14 +38,11 @@ def main():
     op = OptionParser(description=__doc__, version=ver)
     op.add_option('-c', '--config', help='location of config file',
                   default='/etc/apel/receiver.cfg')
-    op.add_option('-l', '--log_config',
-                  help='location of logging config file (optional)',
```
tofu-rocketry, line 42: Add "DEPRECATED" to start of help message.

bin/receiver.py
```diff
@@ -59,7 +56,7 @@ def main():
         print('Cannot start SSM.  Pidfile %s already exists.' % pidfile)
         sys.exit(1)
 
-    ssm.agents.logging_helper(cp, options.log_config)
```
tofu-rocketry, line 62: Restore this line as we'll need to pass `log_config` on for now.

bin/sender.py
```diff
@@ -36,15 +36,13 @@ def main():
     op = OptionParser(description=__doc__, version=ver)
     op.add_option('-c', '--config', help='location of config file',
                   default='/etc/apel/sender.cfg')
-    op.add_option('-l', '--log_config',
-                        help='location of logging config file (optional)',
```
tofu-rocketry, line 40: Add "DEPRECATED" to start of help message.

ssm/agents.py
```diff
@@ -53,15 +53,14 @@
 REFRESH_DNS = 600
 
 
-def logging_helper(cp, log_manual_path=''):
+def logging_helper(cp):
     """Take config parser object and set up root logger."""
     try:
-        if os.path.exists(log_manual_path):
```
tofu-rocketry, line 59: Re-inlcude this line, but add `logging.warning` message, but also note that it needs to be after line 63 where the logging is set up.

tofu-rocketry (changes requested):

tofu-rocketry (approved):

bin/receiver.py
gregcorbett: does this line need to be restored? It looks like it wasn't but the conversation was resolved.

bin/sender.py
```diff
@@ -37,14 +38,20 @@ def main():
...
+
+    # Deprecating functionality.
+    old_log_config_default_path = '/etc/apel/logging.cfg'
+    if (os.path.exists(old_log_config_default_path) or options.log_config is not None):
+        logging.warning('Separate logging config file option has been deprecated.')
 
     cp = ConfigParser.ConfigParser({'use_ssl': 'true'})
     cp.read(options.config)
 
-    ssm.agents.logging_helper(cp, options.log_config)
```
gregcorbett, line 47: As with the receiver, does this line need to be restored?

tofu-rocketry: Path check added to sender.py and receiver.py so we don't need to pass through the path in log_config.

bin/receiver.py
tofu-rocketry: Path check added to sender.py and receiver.py so we don't need to pass through the path in log_config.

PR #217, opened 2022-11-17
Add check that config file exists

rowan04:
- Gives a clearer error message when config file doesn't exist
- Added default paths to the help output on command line for easier use


tofu-rocketry:
Squashed and rebased.

PR #228, opened 2023-03-28
[GT-152] timeout causes ssm to crash

### LINKED ISSUE #214: Read timeout causes SSM to crash
```
2022-10-15 14:00:04,361 - ssm.ssm2 - INFO - Message saved to incoming queue as 0004d08e/634abce458401e
2022-10-15 14:29:12,935 - ssmreceive - ERROR - Unexpected exception: HTTPSConnectionPool(host='msg.argo.grnet.gr', port=443): Read timed out. (read timeout=None)
2022-10-15 14:29:12,935 - ssmreceive - ERROR - Exception type: <class 'requests.exceptions.ReadTimeout'>
2022-10-15 14:29:12,936 - ssmreceive - ERROR - The SSM will exit.
2022-10-15 14:29:12,962 - ssmreceive - INFO - Receiving SSM has shut down.
2022-10-15 14:29:12,962 - ssmreceive - INFO - ========================================
```

Sae126V:
When we make use of `AmsHttpRequests`  provided by the AMS itself, `pull_sub.` is a POST request and it under the hood uses `_retry_make_request()` where it tries to make an attempt for retries. 

In our case, retry =3:
If enabled (retry > 0), multiple subscription pulls will be tried in case of problems/glitches with the AMS service. 

Because we are using the `ArgoMessagingService` class under the hood, it is making use of `requests`  lib. So timeouts (one of the API error status, requests.exceptions.ReadTimeout) from AMS (HTTP 408) or load balancer (HTTP 408 and 504) should be handled by `AmsBalancerException,` `AmsConnectionException,` `AmsTimeoutException.`

Resolves #214 .

### DIFF
--- a/ssm/agents.py
+++ b/ssm/agents.py
@@ -37,7 +37,8 @@ except ImportError:
 
 from stomp.exception import NotConnectedException
 try:
-    from argo_ams_library import AmsConnectionException
+    from argo_ams_library import (AmsConnectionException, AmsTimeoutException,
+                                  AmsBalancerException)
 except ImportError:
     # ImportError is raised when Ssm2 initialised if AMS is requested but lib
     # not installed.
@@ -328,7 +329,9 @@ def run_receiver(protocol, brokers, project, token, cp, log, dn_file):
                     if protocol == Ssm2.STOMP_MESSAGING:
                         ssm.send_ping()
 
-            except (NotConnectedException, AmsConnectionException) as error:
+            except (NotConnectedException, AmsConnectionException,
+                    AmsTimeoutException, AmsBalancerException) as error:
+
                 log.warning('Connection lost.')
                 log.debug(error)
                 ssm.shutdown()


ssm/agents.py
```diff
@@ -329,11 +328,12 @@ def run_receiver(protocol, brokers, project, token, cp, log, dn_file):
...
                         ssm.send_ping()
 
-            except ( 
+            except (
                 NotConnectedException,
                 AmsConnectionException,
-                requests.Timeout,
-                requests.ConnectionError) as error:
+                TimeoutError,
+                ConnectionError
```
gregcorbett, line 335: Is python able to find these classes without the explicit reference to the requests module?

Sae126V: The requests can handle timeout errors or connection(network) related errors. But, It seems we are not leveraging the features of it fully.

TimeoutError and ConnectionError are built-in standard libraries in Python 3. It seems VSCODE extensions came to the rescue when I used python2.7. 

built-in libs


 So. AMS offers an exception to handle timeouts by itself. It should be working fine now, with AMS API timeout exceptions dealt with by the lib.

Sae126V: AMS has these exceptions to handle errors:

class AmsException(Exception):
class AmsServiceException(AmsException):
class AmsBalancerException(AmsServiceException):
class AmsTimeoutException(AmsServiceException):
class AmsConnectionException(AmsException):
class AmsMessageException(AmsException):

Sae126V: @gregcorbett , @tofu-rocketry .. Just wanted to let you guys know that I have updated the description.

tofu-rocketry:
> So timeouts ... should be handled by `AmsBalancerException`, `AmsConnectionException`, `AmsTimeoutException`.

But this clearly isn't the case if you look at the log in #214.  The exception type raised is `requests.exceptions.ReadTimeout`. This won't be a subclass of any of the AMS exception types, right? So trying to catch an AMS exception type won't help. We need to catch the `ReadTimeout`.

Sae126V:
Note:

It seems there is an issue with older version of AMS when `ReadTimeout` Exception occurs. But, Newer versions was able to catch these errors. So, We have decided to catch the issues with what New version of ARGO AMS library is offering. 

For ref:
https://github.com/ARGOeu/argo-ams-library/releases/tag/v.0.6.0-1

ssm/agents.py
```diff
@@ -328,7 +329,13 @@ def run_receiver(protocol, brokers, project, token, cp, log, dn_file):
                     if protocol == Ssm2.STOMP_MESSAGING:
                         ssm.send_ping()
 
-            except (NotConnectedException, AmsConnectionException) as error:
+            except (
```
tofu-rocketry, line 332: The issue with this indentation is that lines 333 to 336 are the same level as 339+.

Could we just go for this style, on two lines, please? (Similar to what you did on line 40.
```
# Aligned with opening delimiter.
foo = long_function_name(var_one, var_two,
                         var_three, var_four)
```

tofu-rocketry (changes requested):
If you address my comment, and rebase the code, then I should be able to slip this into version 3.3.0. 👍🏻

ssm/agents.py
Sae126V: Done with addressing the issue. And, Applied the rebase.

tofu-rocketry (approved):

tofu-rocketry:
I've double checked that these exceptions are available in the AMS lib 0.5.4 and 0.5.1, as used on apel-preprod and apel prod, respectively.

tofu-rocketry:
@Sae126V Can you check that you've merged the latest changes from `dev` into your local copy and rebase again, please?

PR #229, opened 2023-04-05
[GT-151] - Action jobs should be disabled in non-upstream repos

### LINKED ISSUE #211: GT-151 Check for any Actions jobs that should be disabled in non-upstream repos
https://docs.github.com/en/actions/using-jobs/using-conditions-to-control-job-execution

Sae126V:
The condition given to `cron` JOB helps us to NOT trigger the event in a fork branches. And, It will only trigger the CRON Job if it is DEV branch.

"Resolves #211"

### DIFF
--- a/.github/workflows/build-pkgs.yml
+++ b/.github/workflows/build-pkgs.yml
@@ -11,6 +11,8 @@ on:
 jobs:
   build-rpm:
     runs-on: ubuntu-latest
+    if: '!github.event.repository.fork'
+
     env:
       # _github_home is the dir used as a volume by the container
       RPMDIR: /home/runner/work/_temp/_github_home/rpmbuild
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -19,6 +19,8 @@ jobs:
   analyze:
     name: Analyze
     runs-on: ubuntu-latest
+    if: '!github.event.repository.fork'
+
     permissions:
       actions: read
       contents: read


gregcorbett:
can you add "Resolves #211" to the original PR comment? That will link the GitHub PR to the GitHub issue.

From the issue, https://docs.github.com/en/actions/using-jobs/using-conditions-to-control-job-execution showed using the syntax `jobs.<job_id>.if` to prevent a job from running unless a condition is met (i.e. `if: github.repository == 'octo-org/octo-repo-prod`).

Does `github.event.repository.fork` mean the event that gets fired when someone forks a repository? Which wouldn't stop the job from running in forks?

I think both codeql-analysis.yml and build-pkgs.yml define what branches or tags the job should run on, so why is `github.ref == 'refs/heads/dev'` needed?

Sae126V:
Done. I have added Resolves 211 in the PR description.

Regarding the `github.event.repository.fork`
1) `github.event` =>The full event webhook payload. You can access individual properties of the event using this context. This object is identical to the webhook payload of the event that triggered the workflow run and is different for each event.

Now, When I came across this `GitHub Event Types - GitHub Events API,` it had a property called `repository.` 

2) `github.event.repository` => (Object) => This gives where the event occurred.

3) This `Repository (REST API)` has a fork attribute(Which will be taken care of by GitHub and return Boolean Value). `fork` helps us to see if the repo is forked or the original repo.

In short, `github.event.repository.fork` can tell us if it is a forked branch or NOT. 

And you are right @gregcorbett `github.ref == 'refs/heads/dev` is not needed. It just there as an additional check. I will remove that now.

gregcorbett:
LGTM, can you squash the commits down into a single commit?

Sae126V:
Applied rebase on #229.

gregcorbett (approved):
LGTM

tofu-rocketry (approved):

tofu-rocketry:
@Sae126V As this is merged (through a rebase) , can you delete your branch in GitHub now?

PR #249, opened 2023-08-01
repeated _save_msg_to_queue procedure until message can be saved. issue 64

### LINKED ISSUE #64: Dirq sometimes failing to write temporary messages
Following on from #62, while exception handling has been added in #63 for the missing `IOError`, these messages will still be lost, but they will be handled more nicely.

`dirq` seems to be sometimes failing when writing out accepted messages to the `temporary` directory it uses.

This could possibly be solved by retrying when adding the message to the accepted queue.

DanielPerkins7:
resolve #64.   i was unsure which part needed to be repeated so i just repeated the whole function until the message can be saved into the queue, im not sure if this is right though.

### DIFF
--- a/ssm/ssm2.py
+++ b/ssm/ssm2.py
@@ -343,7 +343,11 @@ class Ssm2(stomp.ConnectionListener):
                 name = self._inq.add({'body': extracted_msg,
                                       'signer': signer,
                                       'empaid': empaid})
-                log.info("Message saved to incoming queue as %s", name)
+                try:
+                    log.info("Message saved to incoming queue as %s", name)
+                except:
+                    _save_msg_to_queue(self, body, empaid)
+                
 
         except (IOError, OSError) as error:
             log.error('Failed to read or write file: %s', error)


tofu-rocketry:
From dev meeting:

- It's the `add` methods that are potentially failing, which then ends up in the `except (IOError, OSError)` error handling.
- Retrying `add` up to 3 times would be good.

ssm/ssm2.py
```diff
@@ -316,9 +316,11 @@ def _handle_msg(self, text):
 
         return message, signer, None
 
+    fails = 0
     def _save_msg_to_queue(self, body, empaid):
         """Extract message contents and add to the accept or reject queue."""
         extracted_msg, signer, err_msg = self._handle_msg(body)
+
```
tofu-rocketry, line 323: Careful with adding extra whitespace.

tofu-rocketry (changes requested):

ssm/ssm2.py
```diff
@@ -343,10 +345,27 @@ def _save_msg_to_queue(self, body, empaid):
                 name = self._inq.add({'body': extracted_msg,
                                       'signer': signer,
                                       'empaid': empaid})
-                log.info("Message saved to incoming queue as %s", name)
+                
+                log.info("Message saved to incoming queue as %s", name)
```
rowan04, line 349:
```suggestion
                log.info("Message saved to incoming queue as %s", name)
```
some trailing whitespace here :)

PR #250, opened 2023-08-01
File check. Issue235

### LINKED ISSUE #235: Message directory queue inconsistentcy with sub-directories
`message_directory.py` will just pass on subfolders (alongside files) in the list of items it passes on when iterated. (And in any count of items in the queue.) It should probably filter out non-files.

There is an oblique reference to this in the `sender.cfg`: https://github.com/apel/ssm/blob/3.2.1-1/conf/sender.cfg#L44

DanielPerkins7:
resolve #235 
I was unsure which files needed to be checked but I think checked the right ones so they are only appended if they are files that exist.

### DIFF
--- a/ssm/message_directory.py
+++ b/ssm/message_directory.py
@@ -110,10 +110,7 @@ class MessageDirectory(object):
                 for (file_name, _mtime) in sorted(unsorted_messages,
                                                   key=lambda tup: tup[1]):
                     # Store the sorted file paths in a class element.
-                    if os.path.isfile(file_name):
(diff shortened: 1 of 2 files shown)


tofu-rocketry:
From dev meeting:

- Could you separate this from the commit for #249?
- Approach looks reasonable. If you could add some unit testing for this, that would be great. @gregcorbett did create a suite for the message directory functionality here: https://github.com/apel/ssm/blob/dev/test/test_message_directory.py

RoseECooper (approved):
looks good to me :)

Issue #225, opened 2023-03-13
Deprecate python-ldap usage


tofu-rocketry:
We don't pull info from BDII any more, but can still set host info manually.

Needs two pull requests:
1. Minor change to add warning to logs that functionality is deprecated.
2. Remove functionality.

For the second part, bits to look at:
- `brokers.py` can go
- The bits that try to get a `bdii_url` in `agents.py can go.
- Other bits?

PR #262, opened 2023-09-18
Nick dev brokers

RedProkofiev:
Adding a warning for deprecation of LDAP according to issue #225

### DIFF
--- a/ssm/brokers.py
+++ b/ssm/brokers.py
@@ -43,6 +43,8 @@ class StompBrokerGetter(object):
     def __init__(self, bdii_url):
         """Set up the LDAP connection and strings which are re-used."""
         # Set up the LDAP connection
+        logging.warning('LDAP is deprecated and will be removed in an upcoming version, ' +
+                        'migrate to AMS.')
         log.debug('Connecting to %s...', bdii_url)
         self._ldap_conn = ldap.initialize(bdii_url)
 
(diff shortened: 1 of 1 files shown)


ssm/brokers.py
```diff
@@ -43,6 +43,8 @@ class StompBrokerGetter(object):
     def __init__(self, bdii_url):
         """Set up the LDAP connection and strings which are re-used."""
         # Set up the LDAP connection
+        logging.warning('LDAP is deprecated and will be removed in an upcoming version, ' +
+                        'migrate to AMS.')
         log.debug('Connecting to %s...', bdii_url)
```
tofu-rocketry, line 48:
```suggestion
        logging.warning('LDAP is deprecated and will be removed in an upcoming version, '
                        'please set host locally in SSM config.')
        log.debug('Connecting to %s...', bdii_url)
```

Use implicit line continuation in parentheses rather than adding a + for concatenation.
Tweaked the message as it's not that we specifically want people to use AMS, just not the LDAP retrieval of the broker.

tofu-rocketry (changes requested):
I've created a suggested change, but also can you rebase this onto the latest changes on `dev` to get rid of the duplicate commits from the other PR?

tofu-rocketry (approved):

PR #263, opened 2023-09-18
Add sys.exit() to both sender and reciever

### LINKED ISSUE #189: ssmsend should return non-zero error code if it fails
It seems ssmsend always returns 0 even if it fails, e.g. in this dummy example , or if it fails for other reasons.
```
2022-03-14 22:10:43,280 - ssmsend - INFO - ========================================
2022-03-14 22:10:43,280 - ssmsend - INFO - Starting sending SSM version 3.2.1.
2022-03-14 22:10:43,280 - ssmsend - INFO - Setting up SSM with protocol: AMS
2022-03-14 22:10:43,280 - ssmsend - INFO - No AMS token provided, using cert/key pair instead.
2022-03-14 22:10:43,281 - ssmsend - INFO - No server certificate supplied.  Will not encrypt messages.
2022-03-14 22:10:43,297 - ssmsend - INFO - Messages will be signed using /C=CA/O=Grid/OU=uvic.ca/CN=kermes-dev APEL
SSM failed to complete successfully.  See log file for details.
2022-03-14 22:10:43,297 - ssmsend - ERROR - Unexpected exception in SSM: No option u'capath' in section: u'certificates'
2022-03-14 22:10:43,297 - ssmsend - ERROR - Exception type: <class 'ConfigParser.NoOptionError'>
2022-03-14 22:10:43,297 - ssmsend - INFO - SSM has shut down.
2022-03-14 22:10:43,297 - ssmsend - INFO - ========================================
[centos@apel-test ~]$ echo $?
0
```

So in order to be notified of  a problem, we have to rely on waiting for the sync records to throw an alert
http://goc-accounting.grid-support.ac.uk/rss/CA-VICTORIA-WESTGRID-T2_Sync.html
and then have that picked up by the operations-portal.egi.eu ROD dashboard, and then hopefully someone on duty will make a GGUS ticket for us, all of which could take several days.

If instead ssmsend would return an error code, sites could more quickly and easily internally detect the failure e.g. in a scriptable way, or in our case it would be completely automatic because our ssmsend runs on kubernetes which monitors process return codes.

RoseECooper:
Resolve #189 
Return 1 instead of 0 when failing due to an exception.

### DIFF
--- a/ssm/agents.py
+++ b/ssm/agents.py
@@ -357,11 +357,11 @@ def run_receiver(protocol, brokers, project, token, cp, log, dn_file):
         ssm.shutdown()
         dc.close()
         receiver_failed = True
-  # Currently won't run the else statement due to the while loop in the reciever
-  # Leaving here in case of future refactoring, but commented out so the unreachable
-  # code isn't flagged by tests
-  # else:
-  #     receiver_failed = False
+    # Currently won't run the else statement due to the while loop in the reciever
+    # Leaving here in case of future refactoring, but commented out so the unreachable
+    # code isn't flagged by tests
+    # else:
+    #   receiver_failed = False
 
     log.info('Receiving SSM has shut down.')
     log.info(LOG_BREAK)


ssm/agents.py
```diff
@@ -250,6 +250,7 @@ def run_sender(protocol, brokers, project, token, cp, log):
         print('SSM failed to complete successfully.  See log file for details.')
         log.error('Unexpected exception in SSM: %s', e)
         log.error('Exception type: %s', e.__class__)
+        sys.exit(1)
```
tofu-rocketry, line 253: Putting the exit here will mean lines 255 to 262 don't get executed. I think what we need to do is set a flag here and then use it to set the exit status at the end of `run_sender` (and equivalent for `run_receiver`)

tofu-rocketry (changes requested):

ssm/agents.py
```diff
@@ -344,15 +349,23 @@ def run_receiver(protocol, brokers, project, token, cp, log, dn_file):
...
     except Exception as e:
         log.error('Unexpected exception: %s', e)
         log.error('Exception type: %s', e.__class__)
         log.error('The SSM will exit.')
         ssm.shutdown()
         dc.close()
+        receiver_failed = True
+    # Currently won't run the else statement due to the while loop in the reciever 
+    # Leaving in in case of future refactoring
+    else:
```
tofu-rocketry, line 362: Yeah, let's comment this out for now.

ssm/agents.py
```diff
@@ -259,6 +262,8 @@ def run_sender(protocol, brokers, project, token, cp, log):
 
     log.info('SSM has shut down.')
     log.info(LOG_BREAK)
+    if sender_failed is True:
```
tofu-rocketry, line 265:
```suggestion
    if sender_failed:
```
Let's use the compact format as it's clear enough (as we chose a good variable name).

ssm/agents.py
```diff
@@ -344,15 +349,23 @@ def run_receiver(protocol, brokers, project, token, cp, log, dn_file):
...
         dc.close()
+        receiver_failed = True
+    # Currently won't run the else statement due to the while loop in the reciever 
+    # Leaving in in case of future refactoring
+    else:
+        receiver_failed = False
 
     log.info('Receiving SSM has shut down.')
     log.info(LOG_BREAK)
+    if receiver_failed is True:
```
tofu-rocketry, line 367:
```suggestion
    if receiver_failed:
```

tofu-rocketry (changes requested):

ssm/agents.py
RoseECooper: Done

ssm/agents.py
RoseECooper: Done

tofu-rocketry (changes requested):
Can you fix the Codeclimate issues? (Other than the line too long, as I'm happy with lines that are <100 chars.)

tofu-rocketry (approved):

PR #267, opened 2023-09-19
Stop empty messages being sent and log warning gracefully

### LINKED ISSUE #261: AMS sending silently ignores empty messages
Following #247, I noticed that we're silently ignoring empty messages.

https://github.com/apel/ssm/blob/0ec0b316e07b835670f1f3f8b4e7288cf2cd2fb9/ssm/ssm2.py#L493

`_send_msg_ams` will return `None` for an empty message and the code here will log as if the message was sent:

https://github.com/apel/ssm/blob/0ec0b316e07b835670f1f3f8b4e7288cf2cd2fb9/ssm/ssm2.py#L495

We should wrap that in an `if...else` and log a different message if the message wasn't sent, perhaps a warning.

RedProkofiev:
Resolves #261 

Empty messages and files render as empty lines through Dirq but a simple len() > 0 check means that empty files will now return None, and some changes to send_all now gracefully handle None input to log a warning.


RedProkofiev:
Codeclimate continually fails due to the extra if statement needed for the log/warning element.  Unsure if possible to fix neatly without a function rewrite, so I've put a truly ugly fix here for review as a reminder to rewrite it if needed.

tofu-rocketry:
You've got one of the LDAP commits in here.

gregcorbett:
> Codeclimate continually fails due to the extra if statement needed for the log/warning element. Unsure if possible to fix neatly without a function rewrite, so I've put a truly ugly fix here for review as a reminder to rewrite it if needed.

I would go for that fix / function re-write. Something like the following would work, and that would avoid the need for `log_string[0] == 'M'`. right? 

```
if argo_id is not None:
    log.info("Sent..."
else:
    log.warning("Message..."
```

RedProkofiev:
it wouldn't work: Argo_ID is out of scope in that context because of the way you're doing the logs: the argo_id is used only for constructing the log string.  So the log string is the only thing passed out of scope
I think it'll require a rewrite to make the warning work neatly.  Also, the if...else statement would still fail complexity requirements under 15, I've tried!!!!

Should I just try and rewrite the whole thing?

RedProkofiev:
> it wouldn't work: Argo_ID is out of scope in that context because of the way you're doing the logs: the argo_id is used only for constructing the log string. So the log string is the only thing passed out of scope I think it'll require a rewrite to make the warning work neatly. Also, the if...else statement would still fail complexity requirements under 15, I've tried!!!!
> 
> Should I just try and rewrite the whole thing?

I just realised I'm being very silly here!  sorry, ignore me, it's totally in scope.  However we're resolving to instead just have three separate log message/warning elements rather than upping the complexity with another if...else block. :)

PR #269, opened 2023-09-26
Create msg preprocessor

### LINKED ISSUE #74: Create message pre-processor for split/join
VAC creates single record messages that SSM then duly sends.

Add option to turn on agglomerator, which should be a separate script as it handle some parsing of the messages.

It should check the message header and agglomerate any messages of the same type until it reaches the per message limit (1000) or a new type of message, then send it on or write it back out depending on whether this is just a pre-processing step or part of the sending.

Possible issues
- What to put in the `empaid` label if we're agglomerating multiple messages?
- Failures in the middle of agglomerating.
- What to write out if saving sent messages? (Guess that should be the agglomerated message.)

See also #50.

tofu-rocketry:
Resolves #74


tofu-rocketry:
VAC no longer seems to be an issue - we're not getting all the little messages from those sites any more.

PR #270, opened 2023-09-29
String/Bytes fix

### LINKED ISSUE #146: bytes vs. text  and python 3
Hello!

I recently installed ssm apel_ssm-3.1.1 with python3.5, but kept getting

`AttributeError: 'bytes' object has no attribute 'encode'` 

all the time.  Turned out those exceptions came from the call to crypto.sign in _send_msg_ams, because 'text' is really a list of bytes.  If I change the line

`to_send = crypto.sign(text, self._cert, self._key`

into

`to_send = crypto.sign(text.decode('utf8'), self._cert, self._key)`

it works.

### LINKED ISSUE #209: Check issue about type (str/bytes) of messages in outgoing queue
When testing the preprocessor for messages (using dirq as path_type), the following error would occur (in the function _send_msg_ams in ssm2) : 'Unexpected exception in SSM: write() argument must be str, not bytes' .
Although the processed/combined messages are added to the new outgoing queue (the one with the combined msgs) as strings, they are in bytes when they are read from queue by the sender. Adding a couple of lines to the start of _send_msg_ams , asking for the messages to be converted to strings after they are read, allowed to send the messages.
This issue does not occur when using 'directory' as path_type. The conversion is not necessary because the messages are already strings when read from the outgoing 'queue'.

RedProkofiev:
Resolves #146
Resolves #209 

The PR adds a new test which sends bytestrings through on_message, and through the test isolated two different issues with bytestring handling:

```
 if 'application/pkcs7-mime' in text or 'application/x-pkcs7-mime' in text:
            try:
                text = crypto.decrypt(text, self._cert, self._key)
            except crypto.CryptoException as e:
                error = 'Failed to decrypt message: %s' % e
                log.error(error)
                return None, None, error
```
In _handle_msg, the string comparison with 'in' would trigger a byte-not-str error,
```
 log.warning("Message rejected: %s", err_msg)
 name = self._rejectq.add({'body': body,
                                          'signer': signer,
                                          'empaid': empaid,
                                          'error': err_msg})
 log.info("Message saved to reject queue as %s", name)
```
In _save_msg_to_queue, errors would be caught by the IO/OS error handler, but removing that showed that the raw bytestring from 'body' was being placed into Dirq which handled it as if it was a string causing a decode error.

Our Dirq schemas suggest that we only want strings in this point.  I added a ducktype test that solves both problems in _save_msg_to_queue.

### DIFF
--- a/ssm/ssm2.py
+++ b/ssm/ssm2.py
@@ -318,6 +318,13 @@ class Ssm2(stomp.ConnectionListener):
 
     def _save_msg_to_queue(self, body, empaid):
         """Extract message contents and add to the accept or reject queue."""
+        try:
+            # if not bytes will fail with "'str' obj has no attribute decode"
+            body = body.decode('utf-8')
+        except (AttributeError):
+            # Message type is something string related
+            pass
+
         extracted_msg, signer, err_msg = self._handle_msg(body)
         try:
             # If the message is empty or the error message is not empty
@@ -332,7 +339,6 @@ class Ssm2(stomp.ConnectionListener):
                     body = extracted_msg
 
                 log.warning("Message rejected: %s", err_msg)
-
                 name = self._rejectq.add({'body': body,
                                           'signer': signer,
                                           'empaid': empaid,
--- a/test/test_ssm.py
+++ b/test/test_ssm.py
@@ -1,3 +1,4 @@
+# -*- coding: utf-8 -*-
 from __future__ import print_function
 
 import os
@@ -92,6 +93,15 @@ class TestSsm(unittest.TestCase):
         test_ssm.on_message({'empa-id': 'ping'}, 'body')
         # Check that msg with ID and no real content doesn't raise exception.
         test_ssm.on_message({'empa-id': '012345'}, 'body')
+
+    def test_str_bytes_outgoing(self):
+        """Test to ensure that outgoing messages are converted from Bytes to Str"""
+        test_ssm = Ssm2(self._brokers, self._msgdir, TEST_CERT_FILE,
+                        self._key_path, dest=self._dest, listen=self._listen)
+
+        message = "Appelle Hippocampéléphantocamélos"
+        byte_message = message.encode()
+        test_ssm.on_message({'empa-id': '012345'}, byte_message)
 
     def test_init_expired_cert(self):
         """Test right exception is thrown creating an SSM with expired cert."""


RedProkofiev:
Can confirm, works with python 2.

tofu-rocketry:
> Can confirm, works with python 2.

Except note the Travis CI failure with Python 2 - `test_ssm.py` needs a Unicode declaration at the top.

ssm/ssm2.py
```diff
@@ -326,13 +333,12 @@ def _save_msg_to_queue(self, body, empaid):
                 if signer is None:  # crypto failed
                     signer = 'Not available.'
                 elif extracted_msg is not None:
-                    # If there is a signer then it was rejected for not being
-                    # in the DNs list, so we can use the extracted msg, which
-                    # allows the msg to be reloaded if needed.
+                        # If there is a signer then it was rejected for not being
+                        # in the DNs list, so we can use the extracted msg, which
+                        # allows the msg to be reloaded if needed.
```
tofu-rocketry, line 338:
```suggestion
                    # If there is a signer then it was rejected for not being
                    # in the DNs list, so we can use the extracted msg, which
                    # allows the msg to be reloaded if needed.
```

tofu-rocketry (changes requested):

tofu-rocketry:
Actually sending the test string via AMS works in 2 & 3.

Receiving doesn't work in 2 (but does in 3).

```
2024-02-07 12:56:17,077 - ssmreceive - ERROR - Unexpected exception: 'ascii' codec can't encode character u'\xe9' in position 275: ordinal not in range(128)
2024-02-07 12:56:17,077 - ssmreceive - ERROR - Exception type: <type 'exceptions.UnicodeEncodeError'>
2024-02-07 12:56:17,077 - ssmreceive - ERROR - The SSM will exit.
```

tofu-rocketry:
Oh, I should test dirq sending rather than just directory, as the former was reported as the issue in #270.

tofu-rocketry:
Ok, adding UTF-8 into the mix here was hiding the issue that the original tickets reported. And testing using `directory` rather than `dirq` worked around the original issue.

The problem is the lack of strict encoding and decoding at the program boundaries. This does also mean that there's a choice to make about codec to use, but need to be careful not to break the APEL loader.

PR #289, opened 2024-01-22
Update comments and options in config files

### LINKED ISSUE #200: Update default configs
Especially as moved to AMS

tofu-rocketry:
Resolves #200 

- Set AMS as default messaging system.
- Add AMS hosts (both prod and devel) as options.
- Emphasise note about the server_cert.
- Add token as commented out option.
- Copy log level comments from sender to receiver conf.

### DIFF
--- a/conf/receiver.cfg
+++ b/conf/receiver.cfg
@@ -9,7 +9,7 @@ protocol: AMS
 #bdii: ldap://lcg-bdii.cern.ch:2170
 #network: PROD
 # Alternatively, 'host' and 'port' can be set manually (with 'bdii' and
-# 'network' commented out). This option MUST be used for AMS.
+# 'network' commented out). The 'host' option MUST be used for AMS.
 host: msg-devel.argo.grnet.gr
 #host: msg.argo.grnet.gr
 #port:
@@ -29,7 +29,7 @@ check_crls: false
 ams_project: accounting
 
 # Destination to which SSM will listen.
-destination: ssm2test
+destination:
 
 # Only use direct token auth with AMS if you've been provided with a token to use.
 #token:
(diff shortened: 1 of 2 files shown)


conf/receiver.cfg
```diff
@@ -25,17 +26,22 @@ check_crls: false
 
 [messaging]
 # If using AMS this is the project that SSM will connect to. Ignored for STOMP.
-ams_project:
+ams_project: accounting
 
 # Destination to which SSM will listen.
-destination: /queue/ssm2test
+destination: ssm2test
```
RedProkofiev, line 32: If ssm2test is a dedicated testing architecture, then this is fine, but the intent is for this to be deliberately changed by the user.  So, 'changeme' or blank would be good.

conf/receiver.cfg
```diff
@@ -25,17 +26,22 @@ check_crls: false
...
+#token:
 
 # Accepted messages will be written to <path>/incoming
 # Rejected messages will be written to <path>/reject
 path: /var/spool/apel
 
 [logging]
 logfile: /var/log/apel/ssmreceive.log
+# Available logging levels:
+# DEBUG, INFO, WARN, ERROR, CRITICAL
```
RedProkofiev, line 44: SSM uses at least debug and info (remember that verbosity PR?)  Would this not disrupt that?

Sae126V (commented):
Looks good to me, if we decide to provide default values

conf/receiver.cfg
tofu-rocketry: Done, though went with the blank option.

RedProkofiev (approved):
All looks fine to me.

Issue #299, opened 2024-02-16
Unicode support


tofu-rocketry:
Following on from 3.4.0 which fixed bytes/string issues for Python 3, we should update the decoding (and encoding if necessary) to use utf-8

PR #300, opened 2024-02-16
Character vs. byte strings fix

### LINKED ISSUE #146: bytes vs. text  and python 3

### LINKED ISSUE #209: Check issue about type (str/bytes) of messages in outgoing queue

tofu-rocketry:
Resolves #146 
Resolves #209 

Note that this is a conservative fix and does not add Unicode support so that the behaviour matches between Python 2 and 3.

Unicode support is tracked in #299.

Tested manually between Python 2 and 3, sending and receiving for each, and using `dirq` and `directory` sending.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,7 +3,7 @@
 argo-ams-library
 certifi<2020.4.5.2  # Used by AMS (via requests), 2020.4.5.2 dropped support for Python 2
 pyopenssl >=19.1.0, <=21.0.0  # 22.0.0 dropped support for Python 2
-cryptography==3.3.0  # Crypto dropped support for Python 2 after 3.3
+cryptography==3.3.2  # Crypto dropped support for Python 2 after 3.3
 stomp.py<5.0.0
 python-daemon<=2.3.0  # 2.3.1 dropped support for Python 2
 python-ldap<3.4.0  # python-ldap-3.4.0 dropped support for Python 2
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,7 +3,7 @@
 argo-ams-library
 certifi<2020.4.5.2  # Used by AMS (via requests), 2020.4.5.2 dropped support for Python 2
 pyopenssl<=21.0.0  # 22.0.0 dropped support for Python 2
-cryptography==3.3.0  # Crypto dropped support for Python 2 after 3.3.  They're now on 41.X...
+cryptography==3.2.0  # Crypto dropped support for Python 2 after 3.3
 stomp.py<5.0.0
 python-daemon<=2.3.0  # 2.3.1 dropped support for Python 2
 python-ldap<3.4.0  # python-ldap-3.4.0 dropped support for Python 2
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,8 +2,8 @@
 
 argo-ams-library
 certifi<2020.4.5.2  # Used by AMS (via requests), 2020.4.5.2 dropped support for Python 2
-pyopenssl<=21.0.0  # 22.0.0 dropped support for Python 2
-cryptography==3.2.0  # Crypto dropped support for Python 2 after 3.3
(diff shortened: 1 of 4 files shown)


requirements.txt
```diff
@@ -3,7 +3,7 @@
 argo-ams-library
 certifi<2020.4.5.2  # Used by AMS (via requests), 2020.4.5.2 dropped support for Python 2
 pyopenssl<=21.0.0  # 22.0.0 dropped support for Python 2
-cryptography==3.3.0  # Crypto dropped support for Python 2 after 3.3.  They're now on 41.X...
+cryptography==3.2.0  # Crypto dropped support for Python 2 after 3.3
```
gregcorbett, line 6: You could go into a bit more detail in the commit about what the compatability issues were.

gregcorbett (approved):
I can't think of a better way forward RE: the docker build than what you propose.

requirements.txt
tofu-rocketry: Ah, yes, it was compatibility with pyopenssl. On Python 2:

```
Traceback (most recent call last):
  File "bin/receiver.py", line 21, in <module>
    import ssm.agents
  File "/root/ssm/ssm/agents.py", line 40, in <module>
    from argo_ams_library import (AmsConnectionException, AmsTimeoutException,
  File "/usr/lib/python2.7/site-packages/argo_ams_library/__init__.py", line 11, in <module>
    from .ams import ArgoMessagingService
  File "/usr/lib/python2.7/site-packages/argo_ams_library/ams.py", line 4, in <module>
    import requests
  File "/usr/lib/python2.7/site-packages/requests/__init__.py", line 95, in <module>
    from urllib3.contrib import pyopenssl
  File "/usr/lib/python2.7/site-packages/urllib3/contrib/pyopenssl.py", line 46, in <module>
    import OpenSSL.SSL
  File "/usr/lib/python2.7/site-packages/OpenSSL/__init__.py", line 8, in <module>
    from OpenSSL import crypto, SSL
  File "/usr/lib/python2.7/site-packages/OpenSSL/SSL.py", line 664, in <module>
    _lib.Cryptography_HAS_TLSEXT_HOSTNAME, "SNI not available"
AttributeError: 'module' object has no attribute 'Cryptography_HAS_TLSEXT_HOSTNAME'
```

tofu-rocketry: Aha, we need pyopenssl>=19.1 https://github.com/pyca/pyopenssl/issues/973#issuecomment-741806876

RedProkofiev:
I'm very confused on the docker stuff, but Greg's suggestion might work?  Mainly just here to log that if it's a PyOpenSSL/crypto interaction of versions, we need to have as late a version of Crypto as possible (3.3.2 with py2 compatibility) and PyOpenSSL (21.0.0) :)

requirements.txt
tofu-rocketry: Sorted.

PR #325, opened 2024-04-19
Fix rpmlint to perform rpm checks in given location

Sae126V:
7952bc5

This commit will allow to run the file in home directory and perform rpmlint checks in an given pwd location(Say, when running the FPM script with -s and -b).

### DIFF
--- a/scripts/ssm-build.sh
+++ b/scripts/ssm-build.sh
@@ -211,5 +211,5 @@ then
 else
     # Check for errors in SPEC and built packages
     # For instance; Given $(dirname /root/rpmb/rpmbuild/source) will output "/root/rpmb/rpmbuild".
-    rpmlint $(dirname $SOURCE_DIR)
+    rpmlint "$(dirname "$SOURCE_DIR")"
 fi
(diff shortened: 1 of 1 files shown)


scripts/ssm-build.sh
```diff
@@ -210,5 +210,8 @@ then
     lintian "$BUILD_DIR"/apel-ssm-service_"${TAG}"_all.deb
 else
     # Check for errors in SPEC and built packages
-    rpmlint ~/rpmbuild
+    # For instance; Given $(dirname /root/rpmb/rpmbuild/source) will output "/root/rpmb/rpmbuild".
+    RPMDIR=$(dirname $SOURCE_DIR)
+
+    rpmlint $RPMDIR
```
tofu-rocketry, line 216: Don't really need to split this into two lines.

tofu-rocketry (changes requested):
Good idea, though see inline comment, and please improve the commit message. You can add the text from the PR description as the body text to the commit message (the summary line is ok).

scripts/ssm-build.sh
Sae126V: Updated in a0718d8497b608e0f560fc3b2c8b2fc53ae3cda4. Done

tofu-rocketry (changes requested):
Thanks for the tweaks. Can you sort the Codacy issues out, please?

tofu-rocketry (approved):

PR #339, opened 2024-05-21
Change `exit()` to `sys.exit()`

tofu-rocketry:
The `exit()` function is from site.py which may not be present depending on command line options. `sys.exit` on the other hand, will be there as it is imported.


norealroots (approved):
Looks good to me, happy to approve.

PR #345, opened 2024-07-03
Update pre-commit hooks

### LINKED ISSUE #344: Update pre-commit to use add more checks (that are only available in Py 3)

tofu-rocketry:
Resolves #344 

Set pre-commit-hooks to last Py 3.6 compatible version and add a couple of extra checks.


gregcorbett (approved):
Looks sensible to me :)

PR #362, opened 2024-09-10
Replace optparse with argparse

### LINKED ISSUE #275: optparse deprecating
`optparse` is deprecating (i.e. no longer developed). Although they use the term "deprecated", which would imply no longer available, it is still useable even in the latest versions of Python 3, for now.

`argparse` is the replacement.

William-Brown5515:
Optparse is deprecated - no longer developed. Mostly a like-for-like replacement, but we use vars on the argparse object to output the values in a similar way to optparse.

The old version option on the constructor is unavailable, and so it has been replaced by adding an argument with the 'version' action.

Resolves #275
Resolves GT-540

### DIFF
--- a/bin/receiver.py
+++ b/bin/receiver.py
@@ -21,10 +21,10 @@ from __future__ import print_function
 import ssm.agents
 from ssm import __version__, LOG_BREAK
 
+from argparse import ArgumentParser
 import logging
 import os
 import sys
-from optparse import OptionParser
 
 import configparser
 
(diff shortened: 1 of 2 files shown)


bin/receiver.py
```diff
@@ -37,35 +37,40 @@ def main():
...
-                  help=('location of the file containing valid DNs, '
-                        'default path: ' + default_dns_location),
-                  default=default_dns_location)
-
-    options, unused_args = op.parse_args()
+    arg_parser = ArgumentParser(description=__doc__)
+
+    arg_parser.add_argument('-c', '--config',
+                            help=('location of config file, '
+                            'default path: %s' % default_conf_location),
```
tofu-rocketry, line 44: Watch the indentation on these implicit string concatenations - the indentation should match the start of the string, not the whole line above, otherwise it looks like another argument. I think there's 3 other places to fix this including in the sender. Codeclimate did pick them up (don't worry about the duplication or long lines): https://codeclimate.com/github/apel/ssm/pull/362

bin/receiver.py
```diff
@@ -37,35 +37,40 @@ def main():
...
+    arg_parser.add_argument('-l', '--log_config',
+                            help='DEPRECATED - location of logging config file (optional)',
+                            default=None)
+    arg_parser.add_argument('-d', '--dn_file',
+                            help=('location of the file containing valid DNs, '
+                            'default path: %s' % default_dns_location),
+                            default=default_dns_location)
+    arg_parser.add_argument('-v', '--version',
+                            action='version',
+                            version='Version: %s' % ver)
```
tofu-rocketry, line 55: I don't think we need the "Version:" bit at the start. `ver` should be sufficient as the user is asking for the version info by invoking this option. See also for sender.

tofu-rocketry (changes requested):

bin/receiver.py
```diff
@@ -37,35 +37,41 @@ def main():
...
+                            help='DEPRECATED - location of logging config '
+                                 'file (optional)',
+                            default=None)
+    arg_parser.add_argument('-d', '--dn_file',
+                            help='location of the file containing valid DNs, '
+                                  'default path: %s' % default_dns_location,
+                            default=default_dns_location)
+    arg_parser.add_argument('-v', '--version',
+                            action='version',
+                            version='ver: %s' % ver)
```
tofu-rocketry, line 56:
```suggestion
                            version=ver)
```

Sorry, I wasn't clear. I meant literally just use `ver` as the value as it's already a string with all the info we need. Same for the other file.

bin/receiver.py
```diff
@@ -37,35 +37,41 @@ def main():
...
-    options, unused_args = op.parse_args()
+    arg_parser = ArgumentParser(description=__doc__)
+
+    arg_parser.add_argument('-c', '--config',
+                            help='location of config file, default path: '
+                                  '%s' % default_conf_location,
+                            default=default_conf_location)
+    arg_parser.add_argument('-l', '--log_config',
+                            help='DEPRECATED - location of logging config '
+                                 'file (optional)',
```
tofu-rocketry, line 48: Should probably take out the "(optional)" as it's deprecated. Same again elsewhere.

bin/sender.py
```diff
@@ -36,28 +36,34 @@ def main():
...
-    op.add_option('-l', '--log_config',
-                  help='DEPRECATED - location of logging config file (optional)',
-                  default=None)
-
-    options, unused_args = op.parse_args()
+    arg_parser = ArgumentParser(description=__doc__)
+
+    arg_parser.add_argument('-c', '--config',
+                            help=('location of config file, default path: '
+                                  '%s' % default_conf_location),
```
tofu-rocketry, line 43:
```suggestion
                            help='location of config file, default path: '
                                  '%s' % default_conf_location,
```

This is the only one using parentheses so should tidy them up as they're not necessary in this case.

tofu-rocketry (changes requested):
Almost there. Just note that it's best to just add commits, rather than amend, while code is under review as it makes it easier to see what's changed between each review.

tofu-rocketry (approved):

PR #366, opened 2024-09-11
Update build and deps

### LINKED ISSUE #216: Set min version of ams library (and others?)
As reported (and resolved) here: https://ggus.eu/index.php?mode=ticket_info&ticket_id=159448

AMS library 0.4.2 no longer works with SSM. We should check what the min version we need is and set that in the dependencies. (0.5.1 and 0.6.0 are fine.)

Other dependencies might also need checking.

### LINKED ISSUE #333: EL9: attempt to use unversioned python
It seems that for some time now, using unversioned `__python` macro [throws an error](https://fedoraproject.org/wiki/Changes/PythonMacroError) - certainly on EL9.

The fix is to add
```
%define __python /usr/bin/python3
```
at the top of the spec file.

Given that python2 is long obsolete, it might not be worth to do a conditional

**Side note**
EL9 build also requires updating the requirements:
```
Requires:	python3-stomppy < 8.0.0, python-ldap < 3.5.0, python-setuptools, openssl
```

tofu-rocketry:
This PR updates a bunch of little things now that we no longer need Python 2 support. Mainly:

- Update README
- Update SPEC file so it stays mostly in step with new build script.
- Update build script to remove Python 2 option.
- Update min AMS library version in setup.py.

Resolves #216 
Resolves #333

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,10 +7,10 @@
 
 Secure STOMP Messenger (SSM) is designed to simply send messages
 using the [STOMP protocol](http://stomp.github.io/) or via the ARGO Messaging Service (AMS).
-Messages are signed and may be encrypted during transit.
+Messages are signed and may additionally be encrypted during transit.
 Persistent queues should be used to guarantee delivery.
 
-SSM is written in Python. Packages are available for RHEL 7, and Ubuntu Trusty.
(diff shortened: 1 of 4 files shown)


gregcorbett (approved):
LGTM

PR #372, opened 2024-09-18
Add extra dependencies for apel ssm service

### LINKED ISSUE #371: Change dependency on daemon package
The FPM build script seems to have added in python3-daemon as a dependency, but this should only be required when running SSM as a daemon/service. This has caused issues for users.

As reported by email:
>   - `apel-client-2.1.0-1.el9.noarch` requires `apel-ssm` >= 3.2.0, but none of the providers can be installed
>   - package `apel-ssm-3.4.1-1.el9.noarch` requires `python3-daemon`, but none of the providers can be installed
>   - cannot install the best candidate for the job
>   - nothing provides `python3.9dist(docutils)` needed by `python3-daemon-2.3.2-1.el9.noarch`
> 
> Looks like apel-ssm requires `python3-daemon` but on our current setup the `python-daemon` installation fails.
> 
> Also on the release note of the 3.3.0 you state that `python-daemon` hard requirement was remove but on 3.4.1 the requirement it's still there.

It should be removed from the base package, but if it's possible to add it to the service/pleaserun packages then it can go there as it would be required when running SSM as a service.

Sae126V:
- Remove  `python3-daemon` dependency from the base package. And, add `python3-daemon` for `apel-ssm-service` as it is required.

Resolves #371 
Resolves [GT-555]

When installing the service: Now  python3-daemon  will be installed as part of the service.
Installing:
 apel-ssm-service                noarch                3.4.1-1.el9                    @commandline                 13 k
Installing dependencies:
 python3-daemon                  noarch                2.3.2-1.el9                    epel                         46 k

[GT-555]: https://stfc.atlassian.net/browse/GT-555?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/scripts/ssm-build.sh
+++ b/scripts/ssm-build.sh
@@ -131,7 +131,6 @@ if [[ ${PY_NUM:0:1} == "3" ]]; then
         --depends python3-pip \
         --depends python3-cryptography \
         --depends python3-openssl \
-        --depends python3-daemon \
         --depends 'python3-stomp' \
         --depends openssl "
 
@@ -142,7 +141,6 @@ if [[ ${PY_NUM:0:1} == "3" ]]; then
         --depends python3-pip \
         --depends python3-cryptography \
         --depends python3-pyOpenSSL \
-        --depends python3-daemon \
         --depends openssl \
         --depends openssl-devel "
     fi
@@ -176,6 +174,7 @@ fpm -s pleaserun -t "$PACK_TYPE" \
 --architecture all \
 --no-auto-depends \
 --depends apel-ssm \
+--depends python3-daemon \
 --package "$BUILD_DIR" \
 /usr/bin/ssmreceive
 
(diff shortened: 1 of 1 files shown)


Sae126V:
I have added `python3-dirq` to 8909fc3abc81339a7b8478ba21b24fda733c87b2 for `apel-ssm-service`.

tofu-rocketry:
> I have added `python3-dirq` to [8909fc3](https://github.com/apel/ssm/commit/8909fc3abc81339a7b8478ba21b24fda733c87b2) for `apel-ssm-service`. Does this make sense ? @tofu-rocketry @gregcorbett

Yeah, that makes some sense if we're requiring the daemon for the receiving service - the receiver has to use `dirq` when receiving as that's what the APEL loader then expects to find.

The non-dirq sending is a convenience for non-APEL client sites who just want to dump files into a directory for sending without worrying about the dirq naming schema.

tofu-rocketry (approved):

PR #382, opened 2024-11-28
Remove default host and tidy comments

### LINKED ISSUE #375: Comment out AMS hosts in config
From here: https://ggus.eu/index.php?mode=ticket_info&ticket_id=168453

The installed sender config uses `msg-devel`, but this can trick sites into sending to devel if they don't carefully check their config on a new installation.

Both options should be commented out and documentation added adjacent to those options to explain the difference.

tofu-rocketry:
Resolves #375 in a slightly different way to suggested.

We don't want people accidentally sending to the production host, but we also don't want people to accidentally send to the devel host if not intended. So remove the default host and add the list of hosts to a comment, explaining the purpose of each.

Also uncomment the port option and clarify its usage. It does nothing for AMS sending, so might as well leave it uncommented.

### DIFF
--- a/conf/sender.cfg
+++ b/conf/sender.cfg
@@ -3,10 +3,11 @@
 protocol: AMS
 
 [broker]
-# 'host' and 'port' must be set manually as LDAP broker search is now removed.
-# 'port' is not used with AMS.
-host: msg-devel.argo.grnet.gr
-# port: 443
+# msg-devel.argo.grnet.gr is only for test data
+# msg.argo.grnet.gr is for production data
+host:
+# 'port' is only used for STOMP sending.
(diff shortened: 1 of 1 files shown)


Sae126V (approved):
Make sense, Adrian. LGTM.

tofu-rocketry:
> -> 'port' is only used for STOMP sending. port: 443
> 
> Just one thing, Is this intentional, Adrian. If so, We might need to make a note to update `common/apel/metaconfig/schema.pan` in AQ under sender_cfg_broker_struct -> port ? string to required.

The port options are already optional in the AQ config for SSM, aren't they? So that's fine, right?

PR #384, opened 2024-12-06
Update to improve the naming convention for rpm and deb builds

Sae126V:
- Updated the code to include python(Major+minor) for the naming of builds generated using the FPM script.

### DIFF
--- a/scripts/ssm-build.sh
+++ b/scripts/ssm-build.sh
@@ -110,12 +110,13 @@ rm -f "$TAR_FILE"
 # Get supplied Python version
 PY_VERSION="$(basename "$PYTHON_ROOT_DIR")"
 PY_NUM=${PY_VERSION#python}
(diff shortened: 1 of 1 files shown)


tofu-rocketry:
So the idea was to make it easy to tell what Python version we need for the sake of the site-packages directory location. I've been down bit of a rabbit hole trying to figure if there's a better way than just naming the RPM - it feels like there should be as we don't really care what version we're on beyond it being Python 3 probably. This would suggest that we should actually package the program to install in a non-python location.

In terms of this specific change, it doesn't quite match the usual rpm schemes I can find. They usually use `python3.X-` as a prefix. And we would need to add an Obsoletes and Provides to enable a smooth transition. https://docs.fedoraproject.org/en-US/packaging-guidelines/#renaming-or-replacing-existing-packages

However, I need to have a think about the overall effect we want to achieve, so will put this on hold for now.

PR #405, opened 2025-07-23
Resolve issue with tmpfolder availability after reboot

### LINKED ISSUE #161: /var/run/apel is not persistent under CentOS 7
On CentOS 7 with systemd, /var/run is a symlink to /run mounted on tmpfs. It starts out empty on reboot, so it is up to the software to populate it as necessary.
The common setting for the PID file under /var/run/apel results in failure on startup of the service because the pid file cannot be created.
The common way to deal with this seems to be to create the missing directories on the fly.

The latest rpm release https://github.com/apel/ssm/tree/3.2.0-1  still 'owns' /var/run/apel.

Sae126V:
Resolves [GT-987]

Resolves #161 

[GT-987]: https://stfc.atlassian.net/browse/GT-987?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Sae126V:
Just need to do few checks before raising the PR for SSM.

setup.py
```diff
@@ -81,6 +81,7 @@ def main():
           scripts=['bin/ssmreceive', 'bin/ssmsend'],
           data_files=[(conf_dir, conf_files),
                       ('/etc/logrotate.d', ['conf/apel-ssm']),
+                      ('/etc/tmpfiles.d', ['scripts/run-ssm.conf']),
```
tofu-rocketry: Install to `/usr/lib/tmpfiles.d`

tofu-rocketry: Name should be as per the rpm.

tofu-rocketry (changes requested):

setup.py
Sae126V: Done. I installed the configuration file in /etc/tmpfiles because it overrides any other file with the same name. Similarly, I noticed that /etc/logrotate.d is located within the /etc directory, which is why I chose to place it there as well -  to ensure it takes precedence over any existing file with the same name.

Sae126V: Done

tofu-rocketry:
Have you managed to test on a Debian system? (i.e. Ubuntu)

Sae126V:
> Have you managed to test on a Debian system? (i.e. Ubuntu)

Yes. Managed to test with one of our images i.e, ubuntu-jammy-22.04-nogui . It is working as expected.

tofu-rocketry (approved):

PR #424, opened 2026-05-13
Add alternative method to fetch AMS version

tofu-rocketry:
pkg_resources is deprecated in Python 3.12. importlib.metadata isn't available in 3.6.


rowan04 (approved):
lgtm but obviously the codacy errors will need resolving/ignoring?

PR #441, opened 2026-08-18
Fix handling of corrupted msgs to avoid crash

tofu-rocketry:
Resolves GT-1642

This adds handling of non-ASCII message bodies, which usually indicate a corrupted message, so that the message can be written out to the reject queue and avoid the receiver crashing. The altered code has been moved within the try...except (IOError, OSError) so that the new dirq add() is protected.

Also add a unit test to cover the new code.


garaimanoj (approved):
LGTM

